Issue #70, opened 2026-07-14
Criterion gate: pseudo-replication makes fail verdicts unreliable, needs process-level redesign


kent-tokyo:
## Diagnosis: current gate is not reliable enough to judge individual regressions

The 2026-07-13 run after PR #69 (which correctly fixed the bootstrap-abort bug) showed
nearly every benchmark — including several in crates this PR never touched
(`chematic-smarts`, `chematic-chem`, `chematic-perception`) — with `baseline_count`
96-100/100 and `effect` uniformly ≈ -0.46 to -0.50. That uniformity across unrelated code
paths is not consistent with per-benchmark regressions; it is the signature of a single
shared-environment difference (CPU frequency scaling, VM steal time, scheduler
contention) affecting one side's entire measurement window.

**Root cause: pseudo-replication.** Each `veridict compare` call currently treats one
Criterion process's ~100 internal samples as 100 independent A/B trials. They are not —
they all share one process's runner state on each side:

```
main Criterion process (1 run)  -> ~100 samples, all sharing that run's runner state
PR   Criterion process (1 run)  -> ~100 samples, all sharing that run's runner state
```

A single environment difference between the two processes gets amplified into an
extreme-looking 96-100/100 win rate when fed to `veridict`'s sign-test as if it were 100
independent trials. `effect ≈ -0.46 to -0.50` does **not** mean "46-50% slower" — sign-test
effect is deviation from a 50% win rate, so this means "candidate lost almost every pair,"
which is exactly what one-sided environment contamination produces.

**Standing policy until this is fixed** (adopted 2026-07-14): the Criterion gate stays
non-required and its `fail` verdicts are **not** used as evidence of a real performance
regression. Do not tune `veridict` thresholds or unregister individual "failing"
benchmarks one at a time in response to this gate — the unit that's broken is the
*measurement*, not any specific benchmark.

## Task: Criterion gate reliability redesign

1. Stop pairing raw Criterion sample indices. Take one Criterion **point estimate** per
   process run as a single observation.
2. Run main/PR alternating order (or ABBA blocks: `main -> PR -> PR -> main`, using the
   geometric mean of each side's 2 runs to cancel slow thermal/frequency drift) at least
   7 times each, producing 7+ independent `{id, baseline, candidate}` veridict records
   per benchmark instead of ~100 pseudo-replicates from one process pair.
3. Two-stage screening to keep CI time bounded:
[...]

PR #114, opened 2026-07-19
ci(calibration): [DO NOT MERGE] +5% synthetic regression for issue #70

kent-tokyo:
Throwaway branch for issue #70 item 7 gate-sensitivity calibration. Injects a measured +5% overhead into `parse_smiles_10mol` to check whether the Criterion gate reliably detects a real 5% regression. Will be closed without merging once enough independent runs are collected. See #70.


kent-tokyo:
Stopping per issue #70 item 7 calibration protocol: this run revealed a structural finding (Stage 1's 3-block sign-test cannot mathematically reach a fail verdict regardless of regression size), not a per-run detection question. See issue #70 for the full writeup. Never intended to merge.

PR #115, opened 2026-07-19
ci(calibration): [DO NOT MERGE] +10% synthetic regression for issue #70

kent-tokyo:
Throwaway branch for issue #70 item 7 gate-sensitivity calibration. Injects a measured +10% overhead into `parse_smiles_10mol` to check whether the Criterion gate reliably detects a real 10% regression. Will be closed without merging once enough independent runs are collected. See #70.


kent-tokyo:
Stopping per issue #70 item 7 calibration protocol: first run already hit the stop condition (inconclusive), and the underlying cause turned out to be structural (see issue #70). Never intended to merge.

PR #116, opened 2026-07-19
ci(calibration): [DO NOT MERGE] contamination-detection experiment for issue #70

kent-tokyo:
Throwaway branch for issue #70 item 7 contamination-detection calibration. Source is byte-identical to main -- only `.github/workflows/bench-pr-gate.yml` changes, injecting a background CPU stressor during the criterion-gate step to check whether the existing same-binary null control correctly flags `environment_contaminated=1` instead of leaking a spurious `any_fail=1`. Will be closed without merging once enough independent runs are collected; the workflow diff will be recorded to #70. See #70.

### DIFF
--- a/.github/workflows/bench-pr-gate.yml
+++ b/.github/workflows/bench-pr-gate.yml
@@ -228,6 +228,22 @@ jobs:
       - name: "Run process-level Criterion gate (issue #70 redesign)"
         run: |
           set +e
+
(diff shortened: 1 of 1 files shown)


kent-tokyo:
Stopping the whole-window contamination experiment -- confirmed not a valid one-sided contamination test (environment_contaminated=0, contamination hit both sides roughly equally). Superseded by the Stage 1 fix work tracked in issue #70. Never intended to merge.

PR #127, opened 2026-07-20
feat(io): add streaming Daylight TDT reader and writer

kent-tokyo:
## Summary (IO-2 of 3, stacked — base is IO-1's branch, not `main`)

Second of the 3-part stacked I/O milestone: `main` → IO-1 (#126) → **IO-2 (this PR)** → IO-3 (MRV). Adds streaming Daylight TDT (Tagged Data) file I/O (`.tdt`) — chematic's counterpart to RDKit's `TDTMolSupplier`/`TDTWriter`.

Mandatory RDKit source audit done first (same pinned commit `8afba32ec539dcb2369bc84549d802aca3f7eb39` as IO-1). Findings documented in `crates/chematic-mol/src/tdt.rs`'s module doc comment.

## Rust core (`chematic-mol::tdt`, new)

- `TdtReaderOptions`/`TdtRecordReader<R: BufRead>`, `TdtWriterOptions`/`TdtRecordWriter<W: Write>`, `TdtError` — reuses `MoleculeRecord` (IO-1's shared record type), including its `coordinates_2d`/`coordinates_3d` fields (unused by the SMILES-table format, exactly what TDT's `2D`/`3D` tags need).

**Four deliberate, documented divergences from RDKit — three fix real bugs the audit/oracle found in RDKit itself, not guessed improvements:**
1. **RDKit's own coordinate-list parser drops the last atom's position** (comma-tokenizer bug, confirmed against a live RDKit run both during the audit and again by this PR's own oracle). Chematic parses the full list correctly.
2. **RDKit's `TDTWriter` hard-codes its name tag as `"NAME"` while `TDTMolSupplier`'s own `nameRecord` defaults to `""`** — a bare RDKit round trip silently loses the name, confirmed empirically. Chematic's reader/writer both default `name_tag` to `Some("NAME")`, fixing the round trip out of the box.
3. **RDKit's malformed-tag recovery has an infinite-loop hazard** — an uncaught exception re-throws on the same record indefinitely if a caller naively retries. Chematic's reader always scans forward to the next record boundary internally before returning `Err`.
4. **Discovered via this PR's own oracle run, not predicted by the initial audit**: RDKit drops the final tag line entirely when a file has no trailing newline. Chematic's `BufRead::read_line`-based reader handles this correctly.

## Python bindings

[...]

PR #126, opened 2026-07-20
feat(io): add streaming SMILES table supplier and writer

kent-tokyo:
## Summary (IO-1 of 3, stacked)

First of a 3-part stacked I/O milestone: `main` → **IO-1 (this PR)** → IO-2 (TDT) → IO-3 (MRV). Adds streaming SMILES table file I/O (`.smi`/`.smiles`/`.csv`/`.tsv`/`.txt`) — chematic's counterpart to RDKit's `SmilesMolSupplier`/`SmilesWriter`.

Mandatory RDKit source audit done first (no guessed behavior): `SmilesMolSupplier`/`SmilesWriter`/`SmilesMolSupplierParams` read at pinned commit `8afba32ec539dcb2369bc84549d802aca3f7eb39` (true resolution of tag `Release_2026_03_4`, independently re-verified via the GitHub tags API). Findings documented in `crates/chematic-mol/src/smiles_table.rs`'s module doc comment.

## Rust core (`chematic-mol::smiles_table`, new)

- `MoleculeRecord` (new, shared record type for record-oriented formats beyond MOL/SDF's own `SdfRecord`) — properties are an **insertion-ordered `Vec`**, not a `HashMap`, so writer output is deterministic.
- `SmilesReaderOptions` / `SmilesRecordReader<R: BufRead>` — streaming, forward-only (no `read_to_string` of the whole file).
- `SmilesWriterOptions` / `SmilesRecordWriter<W: Write>`.
- `Delimiter`: `Whitespace` | `Tab` | `Comma` (RFC 4180-subset quoting) | `Custom(u8)`.
- Every row returns `Result<MoleculeRecord, SmilesTableError>` with line/record-index/column context — no silent drops. `strict_parsing` (chematic addition; RDKit's own `SmilesMolSupplier` has no such toggle) controls whether the reader stops or continues after a bad row.
- Promotes `chematic-smiles` from a dev-dependency to a real dependency of `chematic-mol` (already documented in CLAUDE.md's crate-layer diagram; publish graph re-verified after the change).

**Deliberate, documented divergences from RDKit** (see the module doc comment for full citations):
- Forward-only streaming, not RDKit's seek-based random-access `SmilesMolSupplier`.
- One consistent delimiter default (`Whitespace`) — RDKit itself has 4 different defaults across its own entry points (v1 C++ ctor, `setData`, v2 core struct, Python binding).
- Explicit strict/lax parsing toggle — RDKit's `SmilesMolSupplier` has none.
[...]

PR #128, opened 2026-07-21
feat(io): add ChemAxon Marvin (MRV) parser and writer

kent-tokyo:
## Summary

IO-3 of the 3-part stacked I/O milestone (SMILES table → TDT → MRV): adds `parse_mrv`/`write_mrv`
(`crates/chematic-mol/src/mrv.rs`) for ChemAxon Marvin `.mrv` files, plus Python bindings
(`MolFromMrvBlock`/`MolFromMrvFile`/`MolToMrvBlock`/`MolToMrvFile`) and a 206-fixture RDKit oracle
validation. Stacks on IO-2 (`feat/io-tdt`, PR #127) per the original spec — **draft, not for merge**.

### Supported feature matrix

`molecule`/`atomArray`/`bondArray`, atom IDs, element/charge/isotope/atom-map, bond order
(single/double/triple/aromatic), 2D/3D coordinates, wedge/dash stereo.

### Explicitly unsupported (returns `MrvError::UnsupportedFeature`, never a silent partial parse)

S-groups, polymers, reactions, multicenter bonds, query atoms/bonds, R-groups, enhanced stereo
groups, embedded/compressed data. RDKit itself *does* support several of these — chematic's port
does not, by explicit scope decision.

### Implementation

Purpose-built, dependency-free, nesting-aware XML tokenizer (not `cml.rs`'s flat line-scanner,
too weak for MRV's nested elements/child text; not an external XML crate, whose default
DTD/entity posture would need independent verification against this module's own security
requirements anyway). `crate::cml::parse_xml_attrs` is reused for tag-attribute extraction.

**Security** (RDKit's own MRV parser has zero XML hardening — confirmed via source audit, so
nothing to compare against): `<!DOCTYPE`/`<!ENTITY` rejected outright (verified against a
billion-laughs-style entity chain and an XXE local-file-reference payload), input byte-size
limit, element-nesting-depth limit, attribute-value-length limit, duplicate-atom-ID detection (a
real robustness improvement over RDKit, which has none), missing-bond-endpoint detection.

### RDKit oracle — 206 fixtures, RDKit 2026.03.3 (pinned commit `8afba32ec539dcb2369bc84549d802aca3f7eb39`)

Methodology (corrected mid-session — see below): never compare chematic's canonicalizer output
against RDKit's canonicalizer output directly. Both sides are re-canonicalized through RDKit
itself before comparing (`A` = RDKit reads the original fixture; `B` = RDKit re-parses chematic's
SMILES output; `B'` = RDKit re-reads chematic's own MRV write). A 0-atom RDKit "success" is never
counted as a match (RDKit's MRV parser can return an empty `RWMol` for malformed input).

| Metric | Result |
|---|---|
| Parse success | 206/206 |
| chematic read→write→read round trip | 206/206 |
| Read parity (RDKit-original vs. RDKit-reread-of-chematic-SMILES) | 125/206 exact match |
| Write parity (RDKit-original vs. RDKit-reread-of-chematic-MRV) | 203/206 exact match |
| **Unexplained (real structural) mismatches** | **0 / 0** |

Every mismatch falls into one of three documented, non-gating buckets (full detail + evidence
in `validation/mrv_io_parity_summary.json` and `validation/README.md`'s IO-3 section):

1. **69 cases: tetrahedral/E-Z stereo lost converting a wedge-bond-encoded MRV molecule to SMILES.**
   Pre-existing gap, **not MRV-specific** — chematic has no converter from wedge/dash bond
   direction + 2D coordinates into `Atom.chirality` (only into a separate CIP R/S field). MRV's
   own read→write→read round trip is unaffected. MOL V2000 has the identical limitation.
2. **9 cases: a pre-existing `chematic-smiles` writer bug**, discovered incidentally via this
   oracle work — `write()` omits the implicit-H count for a bracket atom forced by
   isotope/charge/atom-map when `hydrogen_count` is `None` (how every non-SMILES format reader
   builds atoms). `[NH4+]` → chematic writes `[N+]`. Confirmed with **zero MRV involvement**
   (`Atom::new(N)` + `charge=1`). Affects every non-SMILES format reader in the workspace, not
   just MRV. MRV's own write path is unaffected (0/9 recur in the MRV-native write comparison).
3. **3 cases: radical-electron info loss on write** (both legs) — `chematic_core::Atom` has no
   radical-electron slot, same documented convention as `mol2000.rs`'s existing
   "doublet radical — treated as neutral."

Findings 1 and 2 are pre-existing, cross-cutting issues unrelated to MRV's own correctness —
flagged for dedicated follow-up, explicitly out of scope for this PR.

**Dedicated kekulize/stereo checks** (independent of the 206-fixture pool): kekulize=True/False
both verified to write the expected bond-order token shape and read back to the same canonical
structure (2/2); `include_stereo=False` correctly drops tetrahedral wedge/dash assignment on
RDKit re-read while preserving connectivity (2/2); `include_stereo` correctly has no effect on
E/Z stereo, which is geometry-derived from always-written 2D coordinates (2/2).

### Adversarial / performance

32 unit/adversarial tests: empty input, deep nesting (depth limit), oversized attribute/input,
billion-laughs entity chain, XXE local-file-reference, truncated input, non-ASCII boundary
content, excessive atom count (20,000), NaN/Infinity coordinates, 500-iteration seeded random
mutation — none panic/hang/OOM. Duplicate IDs, unknown atom refs, unknown elements covered by
dedicated unit tests. ~18,000 documents/sec parsing the 206-fixture pool (informational only).

### Zero regression

Full workspace test suite (`cargo test --workspace --exclude chematic-py --lib --tests`, 219+222
passed), Python (`pytest`, 524 passed — same 2 pre-existing unrelated SMIRKS E/Z failures as
IO-1/IO-2, confirmed present before this PR too), `cargo clippy` (workspace + chematic-py, zero
warnings), `cargo fmt --check`, `cargo deny`, publish graph, `bash scripts/check.sh` all green.

### Python API

```python
from chematic import rdkit_compat as Chem

mol = Chem.MolFromMrvBlock(mrv_text)
block = Chem.MolToMrvBlock(mol)                       # kekulize=True by default, matching RDKit
block = Chem.MolToMrvBlock(mol, kekulize=False)        # order="A" instead of alternating 1/2
Chem.MolToMrvFile(mol, "out.mrv")

# coordinate-preserving round trip (plain Mol wrapper has no conformer slot)
mol, coords_2d, coords_3d = chematic.from_mrv_block_with_coords(mrv_text)
new_block = mol.to_mrv_block_with_coords(coords_2d, coords_3d)
```
`confId != -1` and `prettyPrint=True` raise `NotImplementedError` (no multi-conformer slot, no
pretty-printer) rather than being silently accepted.

### CI note

This repo's own workflow files (`.github/workflows/*.yml`) scope `on: pull_request: branches:
[main]` — stacked PRs (base ≠ `main`) get zero GitHub Actions runs. This is a repo-specific config
choice, not a general GitHub limitation. `bash scripts/check.sh` is the equivalent local gate for
this PR (and was for #127). Changing that CI scope is a separate, explicitly out-of-scope task.

### Status of the 3-PR stack

- IO-1 (#126, `feat/io-smiles-supplier-writer`, base `main`): draft, open.
- IO-2 (#127, `feat/io-tdt`, base `feat/io-smiles-supplier-writer`): draft, open.
- IO-3 (this PR, `feat/io-mrv`, base `feat/io-tdt`): draft, open.

**None of the 3 are merged. This PR stops as a draft per the original instruction — no merge
without new, explicit authorization.**

### DIFF
--- a/crates/chematic-mol/examples/mrv_benchmark.rs
+++ b/crates/chematic-mol/examples/mrv_benchmark.rs
@@ -1,0 +1,56 @@
+//! IO-3 performance record (not a gate, not a Criterion-registered
+//! benchmark -- see `feedback_criterion_gate_pseudo_replication` and the
+//! precedent of `tdt_benchmark.rs`/`smiles_table_benchmark.rs`). Performance
+//! here is purely informational.
+//!
+//! Unlike SMILES-table/TDT (one file, many records), MRV is one document
+//! per molecule -- this reports documents/sec parsing every `.mrv` file in
+//! a directory (e.g. the 206-fixture pool from `gen_rdkit_mrv_oracle.py`).
+//!
+//! Usage:
+//! ```text
+//! cargo run -p chematic-mol --release --example mrv_benchmark -- <fixtures_dir>
+//! ```
+
+use chematic_mol::parse_mrv;
+use std::time::Instant;
+
+fn main() {
+    let dir = std::env::args()
+        .nth(1)
+        .unwrap_or_else(|| panic!("usage: mrv_benchmark <fixtures_dir>"));
+
+    let mut paths: Vec<_> = std::fs::read_dir(&dir)
+        .unwrap_or_else(|e| panic!("read_dir {dir}: {e}"))
+        .filter_map(|e| e.ok())
+        .map(|e| e.path())
+        .filter(|p| p.extension().is_some_and(|ext| ext == "mrv"))
+        .collect();
+    paths.sort();
+
+    if paths.is_empty() {
+        panic!("no .mrv files found in {dir}");
+    }
+
+    let texts: Vec<String> = paths
+        .iter()
+        .map(|p| std::fs::read_to_string(p).unwrap_or_else(|e| panic!("read {p:?}: {e}")))
+        .collect();
+
+    let start = Instant::now();
+    let mut success = 0usize;
+    let mut errors = 0usize;
+    for text in &texts {
+        match parse_mrv(text) {
+            Ok(_) => success += 1,
+            Err(_) => errors += 1,
+        }
+    }
+    let elapsed = start.elapsed();
+    let total = success + errors;
+    let docs_per_sec = total as f64 / elapsed.as_secs_f64();
+
+    println!(
+        "total={total} success={success} errors={errors} elapsed={elapsed:?} docs_per_sec={docs_per_sec:.0}"
+    );
+}
--- a/crates/chematic-mol/examples/mrv_dump.rs
+++ b/crates/chematic-mol/examples/mrv_dump.rs
@@ -1,0 +1,162 @@
+//! IO-3 acceptance-gate dump: parses every fixture named in
+//! `scripts/gen_rdkit_mrv_oracle.py`'s manifest with chematic's own
+//! `parse_mrv`, dumps each fixture's extracted atoms/bonds/coordinates plus
+//! chematic's own isomeric SMILES output as JSONL for
+//! `scripts/mrv_io_parity.py`, which re-canonicalizes both chematic's and
+//! RDKit's output through RDKit itself (never a direct chematic-vs-RDKit
+//! canonicalizer string comparison -- ring-closure digits, atom traversal
+//! order, and branch representation legitimately differ between
+//! canonicalizers for the same graph).
+//!
+//! Also writes each parsed molecule back out via `write_mrv` into
+//! `<out_dir>/chematic_written/<id>.mrv` (chematic-write -> RDKit-read leg),
+//! and performs a chematic-only parse -> write -> parse round trip,
+//! reporting per-fixture identity (chematic-write -> chematic-read leg,
+//! entirely independent of RDKit).
+//!
+//! Usage:
+//! ```text
+//! cargo run -p chematic-mol --release --example mrv_dump -- \
+//!     <manifest.json> <fixtures_dir> <out.jsonl> <written_out_dir>
+//! ```
+
+use chematic_mol::{MrvWriteOptions, parse_mrv, write_mrv};
+use serde_json::{Value, json};
+use std::fs;
+use std::io::Write;
+
+fn atom_json(mol: &chematic_core::Molecule) -> Vec<Value> {
+    mol.atoms()
+        .map(|(_, a)| {
+            json!({
+                "symbol": a.element.symbol(),
+                "charge": a.charge,
+                "isotope": a.isotope,
+                "atom_map": a.atom_map,
+                "aromatic": a.aromatic,
+            })
+        })
+        .collect()
+}
+
+fn bond_json(mol: &chematic_core::Molecule) -> Vec<Value> {
+    mol.bonds()
+        .map(|(_, b)| {
+            json!({
+                "begin": b.atom1.0,
+                "end": b.atom2.0,
+                "order": format!("{:?}", b.order),
+            })
+        })
+        .collect()
+}
+
+fn main() {
+    let args: Vec<String> = std::env::args().collect();
+    let manifest_path = args
+        .get(1)
+        .expect("usage: mrv_dump <manifest.json> <fixtures_dir> <out.jsonl> <written_out_dir>");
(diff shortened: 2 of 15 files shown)


kent-tokyo:
Rebased onto current main (via `git rebase --onto main <old feat/io-tdt tip>`, isolating exactly this PR's 5 MRV-specific commits) after PR #201 (the TDT replacement for the accidentally-closed #127, see that PR's notes) merged. New head: `67a6af2`. Base changed from `feat/io-tdt` (stacked) to `main` directly.

**Full re-classification of the 206-fixture stereo/structural comparison, regenerated completely from scratch** (fresh `.mrv` fixtures re-authored by a live RDKit oracle, not reused) in an isolated `rdkit==2026.03.3`-pinned venv, per this task's explicit instruction not to reuse the old 125/206 / 203/206 numbers:

| Metric | Old (stale base) | **Fresh (this rebase)** |
|---|---|---|
| Read parity (RDKit-original vs. RDKit-reread-of-chematic-SMILES) | 125/206 | **134/206** |
| Write parity (RDKit-original vs. RDKit-reread-of-chematic-MRV) | 203/206 | **203/206** (unchanged) |
| Unexplained/real structural mismatches | 0 | **0** |
| Chematic round-trip (parse→write→parse) | 206/206 | **206/206** (unchanged) |

**Read parity genuinely improved (+9), but investigated the actual mechanism rather than assuming my own pre-rebase hypothesis was right** — it wasn't, in an important way:

- I had flagged, before starting this rebase, that PR #154 (2D wedge/hash→`Atom.chirality` wiring, already on main) postdates this stack's original base and might resolve some of the previously-reported 69 "tetrahedral/E-Z stereo lost" cases. **Checked directly: it does not.** `crates/chematic-mol/src/mrv.rs` has zero references to any `chematic_perception` stereo function (confirmed via `grep -n "use chematic_perception" crates/chematic-mol/src/mrv.rs` — no match, vs. `mol2000.rs` and `cdxml.rs`, which both import and call it). MRV's parser is a separate implementation from MOL/SDF's, and never wires in tetrahedral-from-wedge or E/Z-from-2D-coordinates perception at all. The "69 tetrahedral_or_ez_stereo_lost_converting_wedge_bonds_to_smiles" count is **exactly unchanged** in the fresh classification (69 -> 69) -- confirming this is a real, still-open, MRV-specific gap, not something rebasing resolves.
- The actual +9 came from a **different, independent fix**: the previously-reported "9 cases: chematic-smiles writer bug" bucket (`[NH4+]` incorrectly written as `[N+]`, described in the original PR as affecting *every* non-SMILES format reader, not MRV-specific) is now **absent entirely** from the fresh classification. Verified directly: `chematic_smiles::write()` on a bare `N` atom with `charge=1` and no explicit `hydrogen_count` now correctly emits `[NH4+]`. This was fixed by some other, unrelated merge in this repo's history between when this PR was drafted and now -- not attributable to anything in this rebase or to PR #154.

**Filed as a new, separate follow-up issue** (not fixed here, matches this task's own instruction not to fold MRV-unrelated issues into this PR, and matches this repo's established convention of not conflating independent root causes): the MRV-specific stereo-perception gap (mrv.rs never calling the shared wedge/E-Z perception functions that mol2000.rs and cdxml.rs already use) -- see issue reference below.

**Re-verified everything else:**
- `cargo test -p chematic-mol --lib`: 240/240 pass.
- `bash scripts/check.sh`: full pass (fmt, clippy `-D warnings`, workspace tests, cargo-deny, publish-graph, version — 0.8.0).
- Python: rebuilt via `maturin develop --release`; full pytest 649 passed, only the 2 already-filed (issue #200) pre-existing unrelated failures.
- The dedicated kekulize/stereo checks (`mrv_kekulize_stereo_check.py`) and 32 adversarial/security tests (XXE, billion-laughs, depth/size limits) were not touched by the read-parity reclassification above and are unaffected.

Ready for independent review before merge.

kent-tokyo:
Follow-up issue filed: #202 (MRV reader missing wedge/hash tetrahedral + E/Z stereo perception wiring that mol2000.rs/cdxml.rs already have).

PR #130, opened 2026-07-21
docs(perception): P1-A0 -- diagnose the 2D-stereo reader integration boundary

kent-tokyo:
## Summary

Diagnosis-only PR. No production code changed (no reader, writer, or `chematic-perception::stereo2d` modified). Independent of `#129` (SMILES bracket-H fix), `#126`/`#127`/`#128` (IO-1/2/3 drafts) -- based directly on `main`, does not stack on any of them.

**Headline finding**: the original hypothesis (`chematic_perception::stereo2d::apply_stereo_from_2d` writes the wrong field, `Atom.cip_code` instead of `Atom.chirality`) is incomplete. Source audit found **no reader in the codebase calls these functions at all** -- fixing the field they write to would fix nothing, since nothing calls them from a reader in the first place. The real open question is an integration-boundary one: *when and under what conditions should a coordinate-bearing reader invoke stereo perception?*

This PR answers that question via:
1. **Reader inventory** -- every coordinate-bearing format (MOL V2000/V3000, SDF, CDXML, CML, plus MRV as a pending/unmerged reference) audited for 2D/3D coords, wedge/dash handling, and whether it calls `chematic_perception`.
2. **Current stereo pathway trace**, verified empirically (not just read) -- includes a concrete, previously-undocumented finding: chematic's SMILES writer currently re-encodes a tetrahedral wedge bond as a meaningless OpenSMILES `/`/`\` directional-bond token (no adjacent double bond), which round-trips through chematic's own parser but is silently discarded by any spec-compliant consumer (verified against RDKit) -- worse than a silent drop, since it looks like something survived.
3. **14 frozen diagnostic fixtures** (`crates/chematic-mol/examples/stereo2d_fixture_dump.rs`), one per required mechanism (3-heavy+implicit-H, 4-neighbors-via-explicit-H, genuinely-4-heavy-atoms, solid/hash wedge, reversed bond-atom-order, multiple stereocenters, negative control, CIP-priority tie, degenerate coordinates, E/Z geometry, terminal alkene, contradictory wedges, coord/atom-count mismatch), cross-checked against RDKit (pinned commit `8afba32ec539dcb2369bc84549d802aca3f7eb39`) via `scripts/stereo2d_diagnosis.py`. **14/14 classified, 0 unexplained, 0 baseline drift.**
4. **RDKit source audit** of its own wedge-to-chirality pipeline (`MolFileParser.cpp`/`Chirality.cpp` at the same pinned commit) -- confirms RDKit runs chirality-tag (parity) perception *unconditionally* and *before* sanitization (deliberately, so H-removal can't destroy the wedge signal first), that this parity step is fully CIP-independent, and that 3-heavy+implicit-H stereocenters are treated as first-class, not an edge case.
5. **10 design-boundary questions answered** (default-parse behavior, opt-in vs unconditional, source of truth, error/warning/unspecified policy, etc.), each grounded in the audit rather than guessed.
6. **Recommended split, parity-first** (matches RDKit's own architecture and this PR's own audit -- an earlier draft had this backwards and was corrected during review): **P1-S1a** (wedge + 2D coords + given neighbor order -> `Atom.chirality` + `stereo_neighbor_order`, zero CIP involved) before **P1-S1b** (that parity + CIP ranking -> `Atom.cip_code` R/S label, built on top of S1a, never gating it -- a CIP-priority tie leaves `chirality` intact and only omits the label). **P1-S2** (E/Z bond directions) is independent of both.

Full report: `docs/stereo2d_reader_integration_rfc.md`.

## Review round 1 fixes (commit f146816)

- S1a/S1b order was backwards in an earlier draft (R/S-first, chirality derived from it) -- corrected to parity-first per above, since an R/S-first design can't produce `chirality` at all on a CIP tie.
- `tetrahedral_4heavy_explicit_h` wasn't actually 4 heavy atoms (3 heavy + 1 explicit H) -- renamed to `tetrahedral_4neighbors_explicit_h`, and a genuinely-4-heavy-atom fixture (`C(F)(Cl)(Br)(I)`, zero H anywhere) added. 13 -> 14 fixtures.
- `scripts/stereo2d_diagnosis.py` rebuilt fail-closed: explicit bucket whitelist (not a `startswith("unexpected")` convention), exact fixture-ID-set check, duplicate-ID check, per-bucket evidence assertions, an RDKit-parse-failure bucket, an RDKit version check, and self-tests of the fail-closed logic itself.

## Review round 2 fix (commit 2ea7bee)

- The bucket whitelist alone let mutually-exclusive outcomes of the same mechanism pass regardless of which actually occurred (e.g. `wedge_atom_order_reversed`'s "agrees with RDKit" / "disagrees" / "chematic only" are all individually whitelisted). Added `EXPECTED_BUCKET_BY_ID`, a frozen fixture-ID -> exact-bucket baseline checked in addition to the whitelist; any drift between them now fails closed too.

## Test plan

- [x] `cargo run -p chematic-mol --example stereo2d_fixture_dump > validation/results/stereo2d_fixture_dump.jsonl` -- reproducible, 14 rows, no panics
- [x] `.venv/bin/python scripts/stereo2d_diagnosis.py` -- reproducible, writes `validation/results/stereo2d_diagnosis_summary.json`, 14/14 classified, 0 unexplained, 0 baseline drift, exit 0
- [x] Fail-closed paths verified empirically by hand-injecting each failure mode and confirming exit code 1, then reverting: duplicate fixture ID, missing fixture ID, unknown mechanism, weakened bucket evidence, per-ID baseline drift
- [x] `cargo test --workspace --lib --quiet` -- no regressions (nothing production changed)
- [x] `bash scripts/check.sh` -- fmt/clippy/tests/deny/version all green
- [x] Zero changes to any reader, writer, or `chematic-perception::stereo2d` module (diagnosis calls only existing public APIs)

Not merging -- opening for review per plan.


kent-tokyo:
Addressed all 3 blocking points from review, in commit f146816. Still diagnosis-only, still not merged.

**1. S1a/S1b order reversed** — the recommended split had R/S computed first and `chirality` derived from it, backwards relative to this report's own RDKit audit (chiral tag/parity is CIP-free and computed first; CIP ranking is separate and later). Fixed §7: S1a is now wedge+coords → `chirality`+`stereo_neighbor_order` (no CIP), S1b is parity+CIP-ranking → `cip_code` label, built on top of S1a and preserving `chirality` on a CIP tie. Design question (h) updated to point at S1a specifically.

**2. `tetrahedral_4heavy_explicit_h` wasn't 4 heavy atoms** — confirmed: 3 heavy (F/Cl/Br) + 1 explicit H. Renamed to `tetrahedral_4neighbors_explicit_h`, and added fixture #3 `tetrahedral_4heavy_no_h` (`C(F)(Cl)(Br)(I)`, zero H anywhere) so the two shapes — "H present as a real atom RDKit's `removeHs` would strip" vs "no H to remove at all" — are reported as separate rows. 13 → 14 fixtures.

**3. Diagnosis script made fail-closed** — replaced the `startswith("unexpected")` convention with an explicit `EXPECTED_BUCKETS` whitelist, added an exact `EXPECTED_FIXTURE_IDS` set check (missing/extra IDs abort), a duplicate-ID check, per-bucket evidence assertions (contradictory-wedge now asserts 2 direction tokens actually appear, coord-mismatch asserts a non-empty result actually came back, degenerate-coordinates checks RDKit's own side too), an `rdkit_parse_failed` bucket for parse failures, an RDKit version check, and 2 self-tests of the fail-closed logic itself. `sys.exit(1)` wired through.

Verified empirically, not just written: I hand-injected a duplicate ID, a missing ID, an unknown mechanism, and weakened contradictory-wedge evidence into the dump and confirmed each produces exit code 1 before reverting. Current state: 14/14 classified, 0 unexplained, exit 0.

`bash scripts/check.sh` green, zero production code touched.

PR #131, opened 2026-07-21
feat(perception): P1-S1a-core -- CIP-independent local parity from wedges

kent-tokyo:
## What this is

`chematic_perception::local_parity_from_wedges` computes `Atom.chirality` +
`stereo_neighbor_order` directly from wedge/hash bonds and 2D coordinates,
using the neighbor order already present in the molecule (`mol.neighbors()`).
**It never calls CIP ranking and never touches `Atom.cip_code`.**

This is the first implementation step following the P1-A0 diagnosis (#130):
S1a (parity, CIP-independent) before S1b (CIP-ranked R/S label on top of it).
A CIP tie must not prevent a molecule from having a known local parity --
the wedge drawing is a complete statement on its own.

## Gate: RDKit chiral tag, not R/S

Per the design established during #130's review, the correctness bar here
is agreement with RDKit's raw `CHI_TETRAHEDRAL_CW`/`CCW` tag -- **not** R/S
label agreement (R/S needs CIP, which this function deliberately never
calls).

The sign convention was **measured**, not derived by analogy to the existing
CIP-based `assign_rs`: built matching MOL V2000 fixtures, confirmed
chematic's `mol.neighbors()` order and RDKit's `GetNeighbors()` order are
identical (both are bond-declaration order, so no permutation translation is
needed), then read off which raw signed-volume sign corresponds to which
RDKit tag. Cross-checked a second, independent way via RDKit's own rooted
non-canonical SMILES output and an odd/even-permutation argument -- the two
derivations agree. Full methodology in
`docs/stereo2d_local_parity_calibration.md`.

The 3-heavy+implicit-H case mirrors RDKit's own
`atomChiralTypeFromBondDirPseudo3D` (read from source at the pinned commit
`8afba32ec539dcb2369bc84549d802aca3f7eb39`): no synthetic 3D position is
invented for the missing H. Chematic's existing `signed_volume` already
reduces to RDKit's exact triple product when the center atom itself (not a
neighbor) is used as the pivot.

## Multiple wedges from the same center are not automatically contradictory

An earlier version of this PR rejected any stereocenter with more than one
wedge/hash bond outright. That was an unverified assumption, caught in
review: PR #130's own frozen fixtures (`tetrahedral_4neighbors_explicit_h`,
`tetrahedral_4heavy_no_h`) each legitimately carry a solid wedge to one
substituent and a hash to a *different* one, and RDKit accepts both
cleanly. The blanket rejection would have silently mis-rejected real, valid
drawings.

Fixed: when two or more bonds from a center are wedged, each is checked in
isolation (the already-calibrated single-wedge formula, with every other
wedge zeroed) and all isolated parities must agree -- using a tri-state
sign (`Some(negative)`/`Some(non-negative)`/`None` for a degenerate isolated
volume, not a bare bool, so a degenerate isolated wedge can't spuriously
"agree" with a real one) -- before the combined (all-real-z-at-once) volume
is trusted. Disagreement, or any isolated volume being itself degenerate,
rejects. Measured against RDKit on 5 dual-wedge fixtures -- not derived from
a same-direction/opposite-direction guess, which a first hypothesis floated
and the very first counterexample falsified. Full table in
`docs/stereo2d_local_parity_calibration.md`.

## Test coverage (all required cases from the P1-A0 follow-up spec)

- 3 heavy + implicit H
- 4 neighbors + explicit H
- true 4-heavy (no H anywhere)
- wedge/hash inversion
- bond atom-order inversion (both 3- and 4-neighbor: even vs odd permutation, verified to flip/preserve sign correctly in each case)
- multiple stereocenters
- CIP-priority tie (still produces chirality -- verified against the CIP-based path genuinely refusing the same fixture)
- missing coordinates -> no assignment
- degenerate/coplanar geometry -> no assignment
- two wedges implying disagreeing local parity -> no assignment (two independent fixtures)
- two wedges implying agreeing local parity, on different bonds -> accepted (ported from #130's real fixture geometry, both the 4-neighbor and 3-heavy+implicit-H code paths)
- one isolated wedge volume degenerate while the other is real -> no assignment
- both isolated wedge volumes degenerate -> no assignment

## Plumbing

Adds `chematic_core::Molecule::set_chirality` (no public setter existed;
mirrors the existing `set_cip_code`).

## Explicitly out of scope (per the agreed subsequent order)

- Not called from any reader's default parse path.
- Does not change the SMILES writer. While validating output through the
  existing canonical writer I re-encountered concrete evidence of its
  already-known standalone-wedge bug (`BondOrder::Up`/`Down` is overloaded
  for wedge depiction *and* SMILES `/`/`\` E/Z markers, so a lone wedge bond
  gets written as a meaningless directional token) -- this reinforces why
  that fix has to land before any reader wires this function in
  automatically, not something to fix in this PR.
- Does not add CIP-ranked R/S labeling (P1-S1b, builds on top of this later).
- Branch forked directly from `main` at `659baca`; does not stack on #129 or #130.

Diagnosis-only siblings (aromaticity, descriptor census, canonical-SMILES
residual, ETKDG gap, accurate-CIP audit, ECFP4 bit-exact matrix) are running
in parallel on separate branches/PRs, all also forked from the same
`main@659baca` and not touching this branch's files.

Not merged -- awaiting review. Originally forked from `main@659baca`, before
#129/#130 landed; see the rebase verification section below for the current
state.

## Rebase verification — 2026-07-22

- Rebasing target: `main@48eb526d46eadd3bd3fc042a4d42c9f162397c9d`
- Previous head: `8fac340acb586138bbaf3cda3bfce39df175bb70`
- Current head: `9e310b9bccd85d506c0e736f6d0d10aff9bbdf0a`
- Range-diff: 3/3 commits unchanged
- Conflicts: none
- Dedicated oracle: N/A (code + inline tests only, no diagnosis-script artifacts on this branch); `cargo test -p chematic-perception --lib stereo2d_local` 19/19 passed
- `bash scripts/check.sh`: passed
- GitHub Actions: CI green / Security green / Bench green

### DIFF
--- a/crates/chematic-core/src/molecule.rs
+++ b/crates/chematic-core/src/molecule.rs
@@ -708,6 +708,11 @@ impl Molecule {
         self.atoms[idx.0 as usize].cip_code = code;
     }
 
+    /// Set the tetrahedral chirality (`@`/`@@`) of atom `idx` in-place.
+    pub fn set_chirality(&mut self, idx: AtomIdx, chirality: crate::atom::Chirality) {
+        self.atoms[idx.0 as usize].chirality = chirality;
+    }
+
     /// Return the enhanced stereo groups attached to this molecule.
     pub fn stereo_groups(&self) -> &[StereoGroup] {
         &self.stereo_groups
--- a/crates/chematic-perception/src/lib.rs
+++ b/crates/chematic-perception/src/lib.rs
@@ -15,6 +15,7 @@ pub mod sssr;
 pub mod stereo_validation;
 
 pub mod stereo2d;
+pub mod stereo2d_local;
 
 pub use aromaticity::{
     AromaticityAlgorithm, AromaticityModel, AtomElectronTrace, ConjugatedComponent,
@@ -53,6 +54,7 @@ pub use stereo2d::{
     StereoAssignment2D, apply_stereo_from_2d, assign_ez_from_2d, assign_stereo_from_2d,
     cip_ez_descriptor,
 };
+pub use stereo2d_local::{apply_local_parity_from_wedges, local_parity_from_wedges};
 
 use chematic_core::{AtomIdx, Molecule};
 
--- a/crates/chematic-perception/src/stereo2d.rs
+++ b/crates/chematic-perception/src/stereo2d.rs
@@ -210,10 +210,10 @@ fn cross2d(vx: f64, vy: f64, ux: f64, uy: f64) -> f64 {
 
 /// 3D point (x, y, z).
 #[derive(Clone, Copy)]
-struct P3 {
-    x: f64,
-    y: f64,
-    z: f64,
+pub(crate) struct P3 {
+    pub(crate) x: f64,
+    pub(crate) y: f64,
+    pub(crate) z: f64,
 }
 
 fn assign_rs(mol: &Molecule, coords: &[(f64, f64)], center: AtomIdx) -> Option<CipCode> {
@@ -262,7 +262,7 @@ fn assign_rs(mol: &Molecule, coords: &[(f64, f64)], center: AtomIdx) -> Option<C
 /// Up (wedge toward viewer) → +1.0
 /// Down (dash away from viewer) → -1.0
 /// Anything else → 0.0
-fn wedge_z(mol: &Molecule, center: AtomIdx, neighbor: AtomIdx) -> f64 {
+pub(crate) fn wedge_z(mol: &Molecule, center: AtomIdx, neighbor: AtomIdx) -> f64 {
     // Check bond center→neighbor
     if let Some((_, bond)) = mol.bond_between(center, neighbor) {
         match bond.order {
@@ -312,7 +312,7 @@ fn rank4(mol: &Molecule, center: AtomIdx, subs: &[AtomIdx; 4]) -> Option<[u8; 4]
 // Signed volume
 // ---------------------------------------------------------------------------
 
-fn signed_volume(p1: P3, p2: P3, p3: P3, p4: P3) -> f64 {
+pub(crate) fn signed_volume(p1: P3, p2: P3, p3: P3, p4: P3) -> f64 {
     let (ax, ay, az) = (p1.x - p4.x, p1.y - p4.y, p1.z - p4.z);
     let (bx, by, bz) = (p2.x - p4.x, p2.y - p4.y, p2.z - p4.z);
     let (cx, cy, cz) = (p3.x - p4.x, p3.y - p4.y, p3.z - p4.z);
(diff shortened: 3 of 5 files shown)


kent-tokyo:
Fixed per review: the blanket "reject if >1 wedge" rule was too strict and was itself an unverified assumption.

PR #130's own frozen fixtures (`tetrahedral_4neighbors_explicit_h`, `tetrahedral_4heavy_no_h`) carry a solid wedge + a hash on two *different* bonds from the same center -- standard, valid notation, and RDKit accepts it cleanly. The old code would have silently mis-rejected this exact real case.

Replaced `has_at_most_one_wedge` with `wedges_agree_4`/`wedges_agree_3`: each wedged bond's implied parity is computed in isolation (every other wedge zeroed, reusing the already-calibrated single-wedge formula) and all isolated parities must agree before the combined volume is trusted. Measured against RDKit on 5 dual-wedge fixtures -- not derived from a same-direction/opposite-direction guess, which a first hypothesis floated and the very first counterexample falsified. Full table in `docs/stereo2d_local_parity_calibration.md`.

Added:
- `valid_dual_wedge_solid_and_hash_on_different_bonds_accepted` -- positive, ported from #130's actual fixture geometry.
- `dual_wedge_disagreeing_parity_rejected` -- second independent negative fixture (RDKit itself warns and returns `CHI_UNSPECIFIED` on this exact geometry).
- `valid_dual_wedge_3heavy_same_direction_accepted` -- positive coverage on the other code path (3-heavy + implicit H).

The pre-existing `contradictory_wedges_no_assignment` test still rejects, now confirmed for the right reason (measured disagreement) rather than "more than one wedge."

`bash scripts/check.sh` green. Still not merged; still not rebased (waiting on #129/#130 to land on main first, per the agreed integration order).

kent-tokyo:
Fixed the isolated-volume tri-state guard per review.

\`wedges_agree_4\`/\`wedges_agree_3\` compared isolated volumes via a bare \`< 0.0\` bool, which coded a degenerate (near-zero) isolated volume as "not negative" -- indistinguishable from a real non-negative sign. Added \`volume_sign(f64) -> Option<bool>\` (\`None\` within \`VOLUME_EPS\`, \`Some(negative)\` otherwise); all isolated signs must now be \`Some\` and equal, or it rejects.

Two new regression fixtures:
- one isolated volume degenerate (constructed so a specific wedge's isolated volume is exactly zero regardless of its sign/magnitude -- apex and the two non-wedged viewed points collinear) while the other wedge gives a clear real sign -> rejected
- all four substituents collinear, so every isolated volume among any wedged pair is exactly zero -> rejected

Both existing valid-dual-wedge accept fixtures still pass. \`bash scripts/check.sh\` green. PR body updated (no stale "any second wedge is contradictory" language left). Still not merged, still not rebased.

PR #143, opened 2026-07-22
fix(perception): make build_molecule_from_model authoritative in both directions (K2)

kent-tokyo:
## OPEN DECISION -- needs explicit sign-off before merge

The task brief listed 6 fixtures (tropylium, imidazolium, pyridinium, pyrylium, tellurophene, phosphole) that should all end up with aromatic flags **matching RDKit**. After this fix:

- **4/6 (tropylium, imidazolium, pyridinium, pyrylium) fully match RDKit** under the default `Huckel` engine — verified via the 40-fixture oracle.
- **2/6 (tellurophene, phosphole) cannot**, under the default engine, without extending it to support Se/Te/P generically — which the task's own hard constraints forbid ("do NOT add a new aromaticity model" / RFC section 6 explicitly scopes Se/Te/P support as separate, opt-in, algorithm-shaped future work). Instead they become **correctly, consistently non-aromatic** (atom and bond flags agree with each other) but still **disagree with RDKit** (RDKit says aromatic).

This is a genuine requirement-vs-constraint conflict, not something I resolved unilaterally: the two options are (a) accept honest demotion for these two as the correct outcome under the stated constraints, or (b) relax the constraint and let this PR (or a follow-up) extend default-engine element support for Se/Te/P. I did (a) and named the new bucket accordingly (`kekulize_succeeds_model_consistently_nonaromatic_disagrees_with_rdkit` — deliberately not "correctly_nonaromatic", since that would overclaim RDKit agreement that doesn't exist). Flagging for your call before merge.

## Branch / scope

Branch: `fix/aromaticity-flag-demotion-k2`, forked from `main` at `b184f2234df7783ba4fcbe3867ee85fa5c486bc4` (verified via `git merge-base --is-ancestor`).

**Files touched:**
- `crates/chematic-perception/src/aromaticity.rs` — the production fix (`build_molecule_from_model` + `ring_pi_electrons` + its diagnostic twin `evaluate_atom_pi_contribution_inner`, kept in sync per the file's own anti-drift design).
- `crates/chematic-perception/examples/aromaticity_flag_demotion_k2_corpus.rs` (new) — corpus-level before/after diagnostic dump.
- `crates/chematic-chem/examples/aromaticity_flag_demotion_k2_descriptor_check.rs` (new) — read-only downstream descriptor cross-check.
- `scripts/aromaticity_rdkit_parity_diagnosis.py` — baseline update (new bucket + 8 fixtures' expected bucket, two new self-tests).
- `scripts/aromaticity_flag_demotion_k2_corpus_diagnosis.py` (new) — corpus before/after analysis.
- `validation/results/*` — regenerated/new artifacts (see below).

**Explicitly out of scope, not touched:** `chematic-core/src/kekulization.rs` (K1's territory), `AromaticityAlgorithm`'s default or variants, `chematic-fp`, `chematic-chem` production code, `chematic-smiles`, `chematic-cip`, `chematic-core/src/element.rs`.

**Stop condition:** all reached — 40-fixture oracle re-run and every bucket migration accounted for, zero regressions on the full test suite, corpus + descriptor + ECFP4 cross-checks run and reported, `bash scripts/check.sh` green, PR opened, not merged.

## Mechanism: before vs after

`build_molecule_from_model` previously only **promoted**:
```rust
if model.is_atom_aromatic(idx) { a.aromatic = true; }  // never demoted
```
Now it assigns the model's verdict in **both directions**. Bond orders are computed first (the formula itself is unchanged — model's verdict when it confirms aromaticity, `bond.order` otherwise); an atom's flag is then `model.is_atom_aromatic(idx) OR (it sits on a bond whose final order is still Aromatic)`. That OR-clause is the atomicity/fallback guarantee the task asked for: it only ever fires when the caller passed in a molecule that still carries a stale, never-Kekulized `Aromatic` bond order for a ring the model can't independently verify — there is no trustworthy Kekule value to demote to in that case, so the atom and its bond are kept consistent with each other (both aromatic) rather than the atom being demoted against a bond order that still literally reads "Aromatic" (a worse, self-contradictory state than the pre-fix one). This mirrors RDKit's own sanitization order (`Kekulize` always runs before `setAromaticity` in `MolOps::sanitizeMol`; empirically confirmed: a molecule RDKit can't Kekulize fails `SanitizeMol` outright rather than partially computing aromaticity — `Chem.MolFromSmiles('c1cccc1', sanitize=False)` + `Chem.SanitizeMol` raises `KekulizeException`, tested directly against the pinned `rdkit==2026.03.3`).

A second, coordinated fix was necessary for the demotion to produce **correct** results, not just consistent ones: `ring_pi_electrons` (and its diagnostic twin) had its own charge-blindness bug, independent of K1's `kekulization.rs` fix — a protonated ring N/O and a cationic ring C were scored as if neutral (e.g. pyridinium's `[nH+]` was treated as pyrrole-type, contributing 2π instead of 1π). Fixed by mirroring K1's own charge-aware rules in this separate Hückel pi-electron-counting layer. Without this, fixing `build_molecule_from_model` alone would only make 4 of the 6 fixtures *consistently but wrongly* non-aromatic.

## 40-fixture oracle bucket migrations (`scripts/aromaticity_rdkit_parity_diagnosis.py`, rdkit==2026.03.3)

| Fixture | Before | After |
|---|---|---|
| tropylium_cation | `kekulize_succeeds_model_disagrees_atom_bond_flags_inconsistent` | `matches_rdkit_exact_kekule` |
| imidazolium | same | `matches_rdkit_exact_kekule` |
| pyridinium | same | `matches_rdkit_exact_kekule` |
| pyrylium | same | `matches_rdkit_exact_kekule` |
| azulene | same | `kekulize_succeeds_model_consistently_nonaromatic_disagrees_with_rdkit` (new bucket) |
| selenophene | same | same new bucket |
| tellurophene | same | same new bucket |
| phosphole | same | same new bucket |

All other 32 fixtures unchanged. `unexplained: 0`, `baseline_drift: 0` after updating `EXPECTED_BUCKET_BY_ID`.

## 5,000-molecule corpus (`scripts/descriptor_census_corpus.smi`, reused from PR #137, not rebuilt)

Two calling conventions, both documented as supported by `apply_aromaticity`'s own doc comment:

- **"raw"** (`apply_aromaticity` called directly on freshly-parsed, un-Kekulized input): **0 atom/bond flag changes across all 5,000 molecules.** This is also what every real production caller I found actually does — `chematic-chem/src/descriptors.rs`'s two `apply_aromaticity` call sites (`tpsa`, `tpsa_per_atom`) and `chematic-py`'s `from_smiles` path never Kekulize first (verified by reading the call sites).
- **"kekulized"** (Kekulize first, matching RDKit's own sanitization order and the 40-fixture oracle's own methodology): **499 atom demotions across 84/5,000 molecules, 0 promotions, 0 bond changes.** Self-consistency improved (84→0 molecules with an inconsistent atom/bond flag pair). **This is a genuine RDKit-agreement regression on this pathway, reported honestly, not netted to zero**: checked all 84 directly against RDKit — every one newly *disagrees* with RDKit (0 newly agree). These are pre-existing Hückel-model false negatives on fused bicyclic N-heteroaromatics — a mix of the already-documented azulene-shaped non-alternant case and other fused diazines I did not root-cause further (out of scope for this PR) — unmasked, not introduced, exactly as `docs/aromaticity_rdkit_parity_rfc.md` section 6 predicted ("fixing #2 alone... is not a pure strictly-additive improvement").

Full numbers: `validation/results/aromaticity_flag_demotion_k2_corpus_summary.json`; the 84 changed molecules' full before/after flags: `validation/results/aromaticity_flag_demotion_k2_corpus_changed_only.jsonl`.

## Downstream cross-checks (read-only, no production code touched in either crate)

- **PR #135's ECFP4 bit-exact matrix (6 fixtures):** unaffected. `rdkit_morgan_ecfp4_experimental` depends on `apply_aromaticity_rdkit_parity_experimental` (`crates/chematic-perception/src/rdkit_parity.rs`), a separate engine this PR does not touch (it clears aromatic flags before Kekulizing, so the promote-only bug never applied to it). Re-ran the dump: byte-identical to the committed one, still 6/6 `verified_bit_exact` (0/33 unclassified in the full matrix).
- **PR #137's descriptor census (14 `depends-on-aromaticity` values):** unaffected on this corpus. Computed all 14 (`hbd_count`, `hba_count`, `tpsa`, `fsp3`, `aromatic_ring_count`, `hba_count_lipinski`, `num_aliphatic_rings`, `num_saturated_rings`, `num_aromatic_heterocycles`, `num_aliphatic_heterocycles`, `num_saturated_heterocycles`, `egan_passes`, `pfizer_3_75_passes`, `tpsa_per_atom`) for the 5,000-molecule corpus plus the 6 fixtures, before and after: byte-identical diff (0 changes), consistent with the "raw" pathway finding above (`count_aromatic_rings` itself also only calls `apply_aromaticity` for genuinely-Kekule input with no pre-existing flags at all, bypassing this fix's code path entirely for the standard aromatic-notation corpus).

## Test results

- `cargo test -p chematic-perception --lib`: 149 passed, 2 pre-existing ignored, 0 failed.
- `cargo test --workspace --lib --quiet`: 0 failures across all 19 crates.
- `bash scripts/check.sh`: all green (fmt, clippy `-D warnings`, full workspace unit+integration tests, cargo-deny, publish graph, version consistency).

## Not done / explicitly out of scope

- Se/Te/P support for the default `Huckel` engine (the open decision above).
- The non-alternant/azulene-shaped Hückel model gap, and the other fused-diazine false negatives found in the 84-molecule corpus set — not root-caused, flagged as future work.
- Anything in `chematic-core/src/kekulization.rs`, `element.rs`, `chematic-fp`, `chematic-chem` production code, `chematic-smiles`, `chematic-cip`.


kent-tokyo:
Splitting per coordinator review: the charge-aware ring_pi_electrons fix is now #144 (fix/aromaticity-charge-aware-pi-electrons-k2a, ready to merge). The build_molecule_from_model demotion fix continues as a fresh K2b PR based on #144's tip, with a required deep cluster diagnosis of the 84-molecule/499-flag corpus regression before it can be considered mergeable. Closing this combined PR in favor of the two split ones.

PR #168, opened 2026-07-26
feat(mol): read/write real 3D coordinates in MOL/SDF, diagnose dimension/stereo conflicts

kent-tokyo:
## Summary

3D Breakthrough Program, Agent A. Wires real 3D coordinate support into the MOL/SDF I/O layer (`chematic-mol`), additively, per `docs/3d_breakthrough_master_plan.md` sections 1a/3.

- **Base SHA**: `bcb45e7` (origin/main, includes Wave 0's master-plan merge #164 and the `distance_geometry_v2` module-stub #166)
- **Head SHA**: `c3725b8` (this branch, rebased onto the base above)
- **Files touched**: `crates/chematic-core/src/coords3d.rs` (new), `crates/chematic-core/src/lib.rs` (+2 lines — Coordinator-ratified, see below), `crates/chematic-mol/src/mol2000.rs`, `crates/chematic-mol/src/mol3000.rs`, `crates/chematic-mol/src/sdf.rs`, `crates/chematic-mol/tests/conformer_3d_io.rs` (new), and **`crates/chematic-mol/src/lib.rs`** (root re-exports — added after independent verification found the gap; `lib.rs` was not on the strict per-agent list but is not Coordinator-only either). Nothing else touched — confirmed via `git diff --stat` against base.

**Update after independent Coordinator verification** (see PR discussion): fixed a mislabeled test-count claim, corrected the `is_finite()` justification, added the root re-exports in `chematic-mol/src/lib.rs`, added a semver-relevant behavior-change note below, and filed #171 for a pre-existing, unrelated SDF-reader bug found (not fixed) during this work. Details in the relevant sections below.

## Root cause of the Z-coordinate-discard bug

- `crates/chematic-mol/src/mol2000.rs` (pre-PR): the atom-line parser read only `atom_line.get(0..10)` (x) and `atom_line.get(10..20)` (y) into `coords: Vec<(f64, f64)>`. Bytes `20..30` (z, MDL Ctfile fixed-width layout) were never read anywhere in the function.
- `crates/chematic-mol/src/mol3000.rs` (pre-PR): the V3000 atom-line parser read `tokens[2]`/`tokens[3]` (x/y) but never `tokens[4]` (z).
- Both silently threw the byte/token away — no error, no field to put it in (`MolReadReport`/`SdfRecord` had no 3D-coordinate field at all), and no way for a caller to tell "this file had a Z I discarded" from "this file never had a Z".

## Architecture: `chematic_core::{Coords3D, Point3}`

Per the master plan's decision 1a plus the Coordinator's follow-up review during this task:
- `Point3 { x, y, z }` and `Coords3D { points: Vec<Point3> }` added to new `crates/chematic-core/src/coords3d.rs`, deliberately mirroring `chematic_3d::coords::{Point3, Coords3D}`'s field names, method names (`new_zeroed`/`get`/`set`/`atom_count`), and panic-on-out-of-range indexing convention, so a later Coordinator bridge PR can do `pub use chematic_core::{Coords3D, Point3};` inside `chematic_3d::coords` with minimal call-site changes.
- Two additions beyond what `chematic_3d::coords`'s type has:
  - `is_finite()` on both `Point3` and `Coords3D`. **Not called by this PR's own reader code** — `mol2000.rs`/`mol3000.rs` validate NaN/Inf with a direct `f64::is_finite()` check on the raw parsed value (`mol2000.rs:504`, `mol3000.rs:334`), before a `Point3`/`Coords3D` is ever constructed, so these methods currently have zero non-test callers. They're provided for future consumers (e.g. Wave 2 code that constructs a `Coords3D` from elsewhere and wants to validate it after the fact), not because this PR's own logic needs them — corrected from an earlier, inaccurate framing of these as load-bearing.
  - `Default`/`PartialEq` derives on `Coords3D` (`chematic_3d::coords::Coords3D` derives only `Debug, Clone`) — needed for this crate's own unit tests, harmless additions for a plain data holder.
- `chematic-3d/src/coords.rs` was read-only reference for this PR — not edited, not touched.
- Zero new dependency edges: `python3 scripts/check_publish_graph.py` passes (18 publishable crates, same safe order as before).

**`chematic-core/src/lib.rs` +2-line deviation — Coordinator-ratified.** I added `pub mod coords3d;` **and** `pub use coords3d::{Coords3D, Point3};` (2 lines, not the literal 1 originally specified), needed for `chematic_core::Coords3D`/`Point3` to resolve at root the way the Coordinator's own bridge-PR sketch (`pub use chematic_core::{Coords3D, Point3};`) assumes. Independently verified and approved: narrow, matches the file's existing re-export pattern, and `chematic-core/src/lib.rs` isn't on the strict Coordinator-only list.

**New: root re-exports added to `crates/chematic-mol/src/lib.rs`.** Independent verification found a real usability gap — `MolReadReport`/`SdfRecord` were already root-exported from `chematic_mol`, but their new field types (`CoordinateDimension`, `GeometryRank`, `Stereo3DDiagnostic`, plus the new writer functions and `ConformerEnsemble`/`read_sdf_conformer_ensembles`) were not, so a caller could read `.coordinate_dimension` but couldn't name the type without reaching into `chematic_mol::mol2000::*` directly. Added the missing `pub use` lines, matching the existing pattern. `chematic-mol/src/lib.rs` was not on the strict per-agent file list (only `mol2000.rs`/`mol3000.rs`/`sdf.rs` were assigned specifically) but isn't Coordinator-only either, so this PR now also touches that file — a 12-line, additive-only re-export change, re-verified (fmt/clippy/tests all still green, see Verification).

## Before / after behavior

| | Before | After |
|---|---|---|
| V2000/V3000 z coordinate | silently discarded | parsed into `MolReadReport.conformer: Option<Coords3D>` |
| Malformed z (NaN/Inf/unparseable) | silently discarded (never read at all) | typed `MolParseError::InvalidAtomLine` |
| Missing/blank z field | n/a | still defaults to `0.0` (unchanged leniency, matches x/y) |
| File's own 2D/3D header claim | never read | `MolReadReport.coordinate_dimension: CoordinateDimension` (line-2 columns 20..22, empirically confirmed against RDKit `2026.03.3`'s own `"...RDKit          3D"` output) |
| Whether coordinates are actually flat/coplanar/3D | not tracked | `MolReadReport.geometry_rank: GeometryRank` (`FlatZero`/`Coplanar`/`ThreeD`/`Indeterminate`) — kept **separate** from `coordinate_dimension` per Coordinator review (a real, physically-flat 3D-embedded molecule like benzene must not be conflated with "every z is exactly 0, probably a mislabeled 2D file") |
| Wedge/hash-declared stereo vs. real 3D geometry | never compared | `Stereo3DDiagnostic::WedgeVs3DParityConflict` (tetrahedral, including the 3-heavy + implicit-H case) |
| Writing a 3D conformer to MOL/SDF | impossible (`write_mol_with_coords` always writes `z=0.0`) | new `write_mol_with_conformer` / `write_mol_v3000_with_conformer` / `write_sdf_record_with_conformer` (never write a wedge/hash field — a real 3D geometry makes 2D wedge notation redundant-at-best) |
| Multiple conformers of one molecule across SDF records | no concept | new `read_sdf_conformer_ensembles` groups same-graph records into `ConformerEnsemble { mol, metadata, conformers: Vec<Coords3D> }` |
| Existing `parse_mol`/`parse_mol_with_coords`/`parse_mol_v3000`/`parse_mol_v3000_with_coords`, `write_mol`/`write_mol_with_coords`/`write_mol_v3000`, `SdfReader`/`SdfRecordReader`/`SdfFileReader` | — | **unchanged signatures, unchanged behavior** — verified by full pre-existing test suite still passing (see Verification) |

## Design notes: how each `CoordinateDimension` x `GeometryRank` case is distinguished (Coordinator review)

The Coordinator flagged that collapsing 3D detection to one boolean was the most dangerous simplification. Final design tracks 3 independent signals — `coordinate_dimension` (header claim), `geometry_rank` (observed shape), and `stereo3d_diagnostics` (declared-vs-geometry stereo agreement) — never merged:

1. Header says 2D, coords have nonzero Z -> `Stereo3DDiagnostic::DeclaredTwoDButNonzeroZ { observed: GeometryRank }` (payload is `Coplanar` or `ThreeD`). Test: `declared_2d_but_nonzero_z_is_diagnosed`.
2. Header says 3D, but all points coplanar (not exactly z=0) -> `DeclaredThreeDButFlat { observed: GeometryRank::Coplanar }`. Test: `geometry_rank_coplanar_for_a_real_but_flat_3d_embedding` (real RDKit-embedded benzene, max plane residual ~2.9e-6 A, independently confirmed in the venv -- genuinely flat molecule, not a bug, but still surfaced distinctly from case 3).
3. Header says 3D, every Z exactly 0 -> `DeclaredThreeDButFlat { observed: GeometryRank::FlatZero }`. Test: `declared_3d_but_every_z_exactly_zero_is_diagnosed_as_flatzero`.
4. Genuine 3D coords, no stereo descriptor at all -> zero `Stereo3DDiagnostic` entries, documented explicitly (on the type and in this PR) as "nothing to check," never "verified correct" -- same discipline as the existing `stereo_diagnostics`/`ez_diagnostics` fields. Test: `ez_fixtures_parse_with_real_conformer_and_no_tetrahedral_false_positive` exercises this (real 3D E/Z fixtures, no wedge anywhere, zero tetrahedral diagnostics).
5. Explicit wedge contradicts 3D geometry -> `WedgeVs3DParityConflict`. Tests: `wedge_disagrees_with_3d_geometry_conflict_fires` (tetrahedral-4), `implicit_h_stereocenter_wedge_disagrees_conflict_fires` (3-heavy + implicit-H), both V2000 and the V3000 CFG equivalent.
6. Wedge + nonzero-Z coexist and **agree** -> nothing emitted (3D geometry is the higher-fidelity signal but is never written back to `Atom.chirality`, which is untouched -- this is stated explicitly in `WedgeVs3DParityConflict`'s doc comment). Tests: `wedge_agrees_with_3d_geometry_for_real_rdkit_output_no_conflict`, `implicit_h_stereocenter_wedge_agrees_with_3d_geometry_l_alanine`, `v3000_wedge_agrees_with_3d_geometry_no_conflict`.

## Stereo-vs-3D verification: scope and the one deliberate omission (E/Z)

- **Tetrahedral (implemented, both 4-explicit-neighbor and 3-heavy+implicit-H)**: `wedge_vs_3d_conflicts` reuses the *exact* CIP-independent sign convention `chematic_perception::stereo2d_local::local_parity_from_wedges` already established (apex = first neighbor in `stereo_neighbor_order` for the 4-explicit case; pivot = the center atom itself, no synthetic H position needed, for the 3-heavy+implicit-H case) -- computed on real 3D coordinates instead of a synthetic wedge-derived z. This required **no CIP dependency and no `chematic-3d` dependency**: it reuses `Atom.chirality`/`Molecule::stereo_neighbor_order`, both already produced by the unmodified 2D pathway, as the "declared" side of the comparison.
- **E/Z-vs-3D verification is deliberately deferred, not implemented.** Rationale: MOL/SDF has no wedge-like declaration channel for E/Z independent of the coordinates themselves (unlike tetrahedral parity, which has a real, separate wedge/hash notation). For a genuinely-3D record, the file's "2D projection" (x,y only) of a real 3D structure is an arbitrary-plane artifact, not an independent declaration -- computing an "E/Z from (x,y)" and diffing it against "E/Z from (x,y,z)" would produce noisy, non-actionable diagnostics on legitimately-correct 3D structures (rotating a real 3D structure can make its xy-projection show either cis or trans with no actual conflict). The one place a real declared E/Z channel exists (`chematic_perception::stereo2d_ez_direction`'s `bond_direction` side-channel, which stores `Up`/`Down` on a chosen carrier bond via an axis-referenced sign convention, `direction_for_up`) is `pub(crate)`-private inside `chematic-perception`, not on my file-ownership list, and reimplementing its sign convention independently in `chematic-mol` risks silent drift between two copies of subtle, CIP-adjacent logic. Scoped out per the task's own permission to document rather than risk a wrong "verified" diagnostic. E/Z fixtures (both E and Z, real RDKit 3D embeddings) are still included in the test suite to confirm they at least parse cleanly and produce zero false-positive tetrahedral diagnostics.
- **V2000 legacy atom-parity field (`sss`, bytes 39..42) not read.** This is the MDL-legacy "atom stereo parity" field some writers populate from 3D geometry directly (as an alternative "declared" 3D-stereo channel to wedges). Not implemented -- RDKit's own writer doesn't reliably populate it in the general case either, and decoding its atom-block-order-dependent convention correctly would be a second, separate research effort. Documented as a known limitation, not silently skipped.
- **Implicit-H stereocenters are covered**, per the task's explicit ask -- not skipped, not approximated with a synthetic H position: the 3-heavy + implicit-H sign convention above needs only the center atom's own real 3D position plus its 3 real heavy-neighbor positions (all present in the conformer), so no virtual-H heuristic was needed at all.

## Typed errors added

All routed through the **existing** `MolParseError` enum (not touched, per file ownership -- `crates/chematic-mol/src/error.rs` is not on my file list):
- `InvalidAtomLine` for a present-but-unparseable z (V2000 and V3000).
- `InvalidAtomLine` for a present-but-non-finite z (NaN, or `Inf` from float-overflow syntax like `1e400` -- valid Rust float *syntax*, not a parse failure by Rust's own rules, but not a finite coordinate).
- A genuinely **missing/blank** z (line too short in V2000, or a blank-but-present 10-column field) still silently defaults to `0.0` -- unchanged, matches x/y's existing leniency. This distinction (missing vs. garbled) was called out explicitly by the pre-work reviewer and is regression-tested (`v2000_blank_z_field_still_defaults_to_zero_not_an_error`).
- Atom-count-mismatch-between-header-and-block: V3000 already had this (`"expected {N} atoms, found {M}"`, pre-existing, unmodified). V2000's format has no independent second signal to cross-check against (the counts line's `natoms` field is the *only* source of truth for how many lines to consume -- there's no `END ATOM` marker as in V3000), so there's nothing new to add there; the existing `UnexpectedEnd` on a too-short file already covers the truncation case.

## Behavior change (semver-relevant -- for Coordinator's CHANGELOG; `CHANGELOG.md` is Coordinator-only, not edited by this PR)

`read_mol_with_diagnostics`/`read_mol_v3000_with_diagnostics` -- and therefore everything downstream that calls them (`parse_mol`, `parse_mol_with_coords`, `parse_mol_v3000`, `parse_mol_v3000_with_coords`, `SdfReader`, `SdfRecordReader`, `SdfFileReader`) -- now return `Err(MolParseError::InvalidAtomLine)` on input that previously parsed successfully: a garbled or non-finite (NaN/Inf) **z**-coordinate field. Before this PR, z was never read at all, so any text in that column silently had no effect. x/y are unaffected and keep their pre-existing lenient-default-to-`0.0` behavior for backward compatibility -- only z gained the new strictness, since nothing downstream ever consumed it before, so there is no legitimate prior behavior to preserve there. Confirmed nothing in the existing pre-PR test corpus (238 tests, see Verification) trips this. Suggested CHANGELOG line:

> **Changed** (semver-minor, additive elsewhere): MOL/SDF readers now reject a garbled or non-finite (NaN/Infinite) Z-coordinate field with a typed `MolParseError::InvalidAtomLine` instead of silently ignoring it (Z was never read/validated before). X/Y coordinates keep their existing lenient defaulting.

## Verification

Independent Python venv (never touched the shared repo `.venv`, confirmed -- see below): scratchpad-local `venv_agentA`, `rdkit==2026.3.3` (pinned commit `8afba32ec539dcb2369bc84549d802aca3f7eb39`, same pin `scripts/stereo2d_diagnosis.py`/`docs/etkdg_3d_gap_rfc.md` use).

**Fixture set** (`crates/chematic-mol/tests/conformer_3d_io.rs`, 27 tests, denominators below):
- 2D-flag+2D-coords, 3D-flag+3D-coords (V2000 and V3000): ethanol, both flags -- real RDKit `AllChem.EmbedMolecule`/`Chem.MolToMolBlock`/`Chem.MolToV3KMolBlock` output.
- 3D-flag+flat(Z=0)-coords / 2D-flag+nonzero-Z: hand-edited (single substring replace of the header's dimensional-code field only, geometry byte-identical to the real RDKit fixture) -- 2/2 tests pass.
- Tetrahedral R/S: `[C@]`/`[C@@](Br)(F)(Cl)I`, real RDKit embeddings, both orientations -- 2/2 self-consistency tests pass, plus 2/2 deliberate-conflict tests (single-digit wedge-field flip, geometry untouched) pass, V2000 and V3000 both.
- Implicit-H R/S: L-alanine / D-alanine, real RDKit embeddings (RDKit's own wedge notation on the alpha-carbon-to-methyl bond, exactly the 3-heavy+implicit-H case) -- 3/3 self-consistency + cross-enantiomer tests pass, plus 1/1 deliberate-conflict test passes.
- E/Z: `Cl/C=C/Br` and `Cl/C=C\Br`, real RDKit embeddings -- parse cleanly, zero false-positive tetrahedral diagnostics (E/Z-vs-3D itself deferred, see above).
- Enhanced-stereo-group case: hand-constructed V3000 block (real 3D coordinates from the bromochlorofluoroiodomethane fixture + a `STEABS` `COLLECTION` block) -- confirms `stereo_groups()` and `conformer` coexist correctly.
- Multiple conformers: `AllChem.EmbedMultipleConfs` (3 conformers of butane) written as 3 separate real SDF records -> `read_sdf_conformer_ensembles` groups them into exactly 1 `ConformerEnsemble` with 3 conformers, denominator 1/1.
- Typed-error fixtures (NaN/Inf/garbled/blank z, V2000 and V3000): 7/7 pass, hand-authored with a byte-exact-column-constructing Python helper (not eyeballed) to avoid off-by-one column bugs.
- Round-trip (write 3D -> reparse): V2000 and V3000, both confirm `coordinate_dimension == ThreeD`, zero new `stereo3d_diagnostics`, coordinates preserved within writer precision (+/-1e-3 A, the `{:.4}` format's rounding).
- **27/27 pass.**

**Corrected test-count framing** (an earlier version of this PR body mislabeled the post-PR total as "pre-existing" -- fixed after independent verification):
- **Pre-existing** (`origin/main`, `chematic-mol` + `chematic-core` `--lib`): **238 tests** (99 core + 139 mol).
- Independently re-confirmed via a `comm`-diff of sorted `--list` test names between `origin/main` and this branch: **0 removed/renamed**, **6 added** (all in `chematic-core`'s new `coords3d::tests::*`).
- **New in this PR**: +6 `chematic-core` lib tests (`coords3d::tests`) + 27 `chematic-mol` integration tests (`conformer_3d_io`, a separate test binary, not part of `--lib`) = **33 new tests**.
- **Total on this branch**: 238 + 33 = **271** across `chematic-mol` + `chematic-core` (`--lib` + `conformer_3d_io`). The `244/244` figure in the command list below is `105 core lib (99+6) + 139 mol lib (unchanged)` -- a real, passing number, just not "pre-existing."

**Commands run** (from repo root):
```bash
cargo fmt --all -- --check                                    # pass
cargo clippy --workspace --all-targets -- -D warnings          # pass (full workspace, ~2 min)
cargo build --workspace --exclude chematic-py                  # pass (chematic-py needs maturin
                                                                 #  to link its PyO3 cdylib; this
                                                                 #  fails identically on unmodified
                                                                 #  origin/main -- confirmed, pre-existing,
                                                                 #  not caused by this PR)
python3 scripts/check_publish_graph.py                          # pass, 18 publishable crates, zero new edges
cargo test -p chematic-mol -p chematic-core --lib --quiet       # 244/244 pass -- see denominators below,
                                                                 #  this total is NOT "244 pre-existing"
cargo test -p chematic-mol --test conformer_3d_io               # 27/27 new integration tests pass
cargo test --workspace --lib --quiet                            # pass, every crate 0 failed
                                                                 #  (this ran slowly due to concurrent
                                                                 #  Wave-1 sibling-agent builds sharing
                                                                 #  the cargo registry/package cache in
                                                                 #  this multi-agent environment, but
                                                                 #  completed clean, exit code 0)
```

**RDKit fixture generation**: SMILES -> `AllChem.EmbedMolecule`/`EmbedMultipleConfs` -> `Chem.MolToMolBlock`/`Chem.MolToV3KMolBlock`, `randomSeed=0xF00D` (same seed convention as `docs/etkdg_3d_gap_rfc.md`), `AllChem.MMFFOptimizeMolecule` post-embed. The generator scripts live in the independent venv's scratch dir, not committed (the fixtures themselves are baked into the Rust test file as string constants).

## Known limitations / deferred (honest, not silent)

1. **E/Z-vs-3D-geometry verification is not implemented** (see rationale above) -- tetrahedral-only for now.
2. **V2000 legacy atom-parity field (`sss`) is not read** -- deferred, documented.
3. ~~No root re-export of the new public items~~ **Fixed**: `CoordinateDimension`, `GeometryRank`, `Stereo3DDiagnostic`, `write_mol_with_conformer`, `write_mol_v3000_with_conformer`, `write_sdf_record_with_conformer`, `ConformerEnsemble`, `read_sdf_conformer_ensembles` are now all re-exported at `chematic_mol::` root (`crates/chematic-mol/src/lib.rs`), matching the existing pattern for `MolReadReport`/`SdfRecord`. `lib.rs` is not on the strict per-agent list but isn't Coordinator-only, so added directly per independent verification's finding.
4. **`same_graph_identity` (the SDF conformer-ensemble grouping check) is deliberately simple**: order-sensitive structural equality (same atom/bond count, same per-index element/charge/isotope, same per-index bond endpoints/order) -- not full graph isomorphism or canonical-SMILES equality. This is correct for the concrete case this program's task described (the same molecule written N times by the same embedding tool, which never reorders atoms/bonds between calls) and is explicitly **not** meant to replace Agent B's more rigorous, parallel identity-layer work; Coordinator is expected to reconcile the two.
5. ~~Found, not fixed: a pre-existing, unrelated bug in the shared SDF-block-splitting helpers~~ **Filed as issue #171**: `read_sdf_with_diagnostics`/`SdfReader`/`SdfRecordReader`'s "skip leading blank lines" preprocessing strips a legitimately-blank *first* line of a record (a molecule with no name, which RDKit's own `Chem.MolToMolBlock` produces routinely), not just stray blank lines between records -- root-caused to `mol2000.rs:715`, `sdf.rs:40`, `sdf.rs:163`, with a confirmed partial historical workaround already in `chematic-py/src/formats.rs:248` (since v0.4.10) that this PR's own new reader inherits the same exposure to. Found incidentally while building the multi-conformer-ensemble fixture; worked around in the one test that would otherwise hit it (`sdf_conformer_ensemble_omits_records_with_no_conformer`, by giving the fixture molecules real names) rather than fixed here, per scope discipline (general SDF delimiter-handling, unrelated to 3D coordinates). See https://github.com/kent-tokyo/chematic/issues/171 for the full repro and proposed fix.
6. **Conformer selection on write is caller-driven, no selector API**: `write_mol_with_conformer`/`write_mol_v3000_with_conformer`/`write_sdf_record_with_conformer` each take a single `&Coords3D`; picking which conformer out of a `ConformerEnsemble.conformers: Vec<Coords3D>` to write is left to the caller (index it themselves) -- deliberately, per this task's own "keep it simple" instruction.
7. **Coordinate epsilons are fixed constants**, not configurable: `Z_EPS = 1e-4` A (flat-vs-nonzero), `COPLANAR_EPS = 1e-2` A (plane-residual, chosen well above float noise and well below a real ring pucker ~0.2-0.5 A), `COLLINEAR_EPS = 1e-6` A^2 (plane-defining-vector-pair search), `VOLUME_EPS = 1e-6` A^3 (degenerate tetrahedron cutoff for the wedge-vs-3D check). Not exposed as parameters.
8. **`classify_geometry_rank`'s plane-fit search fixes the origin at point 0** (O(n) instead of O(n^3)) -- misses only the vanishingly rare pathological case where point 0 happens to be collinear with every other point yet some other, unrelated triple isn't. Documented at the call site; not expected to matter for real molecular input.

## Confirmation: shared `.venv` never touched

All RDKit oracle work ran through an independent venv created fresh in the scratchpad directory (`python3 -m venv`, `pip install rdkit==2026.3.3`), never inside the repo. No `maturin develop`, `pip install -e`, or `pytest` was run against the shared repo `.venv` at any point in this task.

## Merge blockers / open questions for the Coordinator

All items raised by independent verification have been addressed on this branch (pushed):
1. Test-count claim corrected (238 pre-existing, 0 removed/renamed, 33 new, 271 total) -- was mislabeled.
2. `is_finite()` justification corrected (not currently called by this PR's own reader code); `Default`/`PartialEq` derives now mentioned.
3. Semver-relevant behavior-change note added (new `Err` on garbled/non-finite z) with a suggested CHANGELOG line for the Coordinator to add to `CHANGELOG.md` (not edited here, per file ownership).
4. `chematic-core/src/lib.rs` +2-line deviation -- ratified, no action needed.
5. Root re-exports added to `crates/chematic-mol/src/lib.rs` for the new public items.
6. Pre-existing SDF blank-name-line bug filed as https://github.com/kent-tokyo/chematic/issues/171 (not fixed here, per scope discipline).

No open questions remain from my side. Not self-merging, per program rules -- this is still a draft PR, timing of the actual merge is the Coordinator's call per the program's dependency order (Coords3D bridge PR).

### DIFF
--- a/crates/chematic-core/src/coords3d.rs
+++ b/crates/chematic-core/src/coords3d.rs
@@ -1,0 +1,228 @@
+//! Minimal, algorithm-free 3D coordinate storage shared across crates.
+//!
+//! [`Coords3D`] holds one [`Point3`] per atom, indexed by atom insertion
+//! order (`AtomIdx.0` as `usize`). It carries no embedding, force-field, or
+//! alignment logic -- see the 3D Breakthrough Program's master plan
+//! (`docs/3d_breakthrough_master_plan.md`, decision 1a) for why this type
+//! lives in `chematic-core` rather than `chematic-mol` importing
+//! `chematic_3d::coords::Coords3D` directly: `chematic-3d` pulls in
+//! `chematic-ff`/`chematic-chem`/`chematic-fp`/`chematic-smarts` transitively,
+//! an unwanted dependency footprint for `chematic-mol` and every other
+//! `chematic-core` consumer that never touches 3D generation. Every crate
+//! already depends on `chematic-core`, so this requires zero new dependency
+//! edges.
+//!
+//! **Field layout and method names deliberately mirror
+//! `chematic_3d::coords::{Point3, Coords3D}` as closely as possible** (same
+//! `x`/`y`/`z` fields, same `points: Vec<Point3>` layout, same
+//! `new_zeroed`/`get`/`set`/`atom_count` names and panics-on-out-of-range
+//! indexing convention) so that a later Coordinator-authored bridge PR can
+//! make `chematic_3d::coords` re-export this type
+//! (`pub use chematic_core::{Coords3D, Point3};`) with minimal changes to
+//! existing `chematic-3d` call sites. `chematic-3d`'s own `Coords3D` is not
+//! touched by this PR (read-only reference); the two types are reconciled at
+//! Wave 1->2 integration time, not here.
+//!
+//! This module adds exactly one thing `chematic_3d::coords` doesn't have:
+//! `is_finite()` on both types, since `chematic-mol`'s readers need a way to
+//! reject NaN/Inf coordinates as a typed parse error rather than silently
+//! constructing a poisoned conformer.
+
+use crate::molecule::AtomIdx;
+
+/// A 3D point in Cartesian space (Angstrom).
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Point3 {
+    pub x: f64,
+    pub y: f64,
+    pub z: f64,
+}
+
+impl Point3 {
+    /// Create a new point.
+    #[inline]
+    pub fn new(x: f64, y: f64, z: f64) -> Self {
+        Self { x, y, z }
+    }
+
+    /// The origin (0, 0, 0).
+    #[inline]
+    pub fn zero() -> Self {
+        Self::new(0.0, 0.0, 0.0)
+    }
+
+    /// `true` when all three components are finite (not NaN, not Infinite).
+    #[inline]
+    pub fn is_finite(&self) -> bool {
+        self.x.is_finite() && self.y.is_finite() && self.z.is_finite()
+    }
+
+    /// Euclidean distance to another point.
+    #[inline]
+    pub fn distance(&self, other: &Self) -> f64 {
+        let dx = self.x - other.x;
+        let dy = self.y - other.y;
+        let dz = self.z - other.z;
+        (dx * dx + dy * dy + dz * dz).sqrt()
+    }
+
+    /// Component-wise addition.
+    #[inline]
+    pub fn add(&self, other: &Self) -> Self {
+        Self::new(self.x + other.x, self.y + other.y, self.z + other.z)
+    }
+
+    /// Component-wise subtraction.
+    #[inline]
+    pub fn sub(&self, other: &Self) -> Self {
+        Self::new(self.x - other.x, self.y - other.y, self.z - other.z)
+    }
+
+    /// Scalar multiplication.
+    #[inline]
+    pub fn scale(&self, s: f64) -> Self {
+        Self::new(self.x * s, self.y * s, self.z * s)
+    }
+
+    /// Dot product.
+    #[inline]
+    pub fn dot(&self, other: &Self) -> f64 {
+        self.x * other.x + self.y * other.y + self.z * other.z
+    }
+
+    /// Cross product.
+    #[inline]
+    pub fn cross(&self, other: &Self) -> Self {
+        Self::new(
+            self.y * other.z - self.z * other.y,
+            self.z * other.x - self.x * other.z,
+            self.x * other.y - self.y * other.x,
+        )
+    }
+
+    /// Euclidean norm (length).
+    #[inline]
+    pub fn norm(&self) -> f64 {
+        (self.x * self.x + self.y * self.y + self.z * self.z).sqrt()
+    }
+
+    /// Normalize to a unit vector.
+    ///
+    /// # Panics
+    /// Panics if the vector has zero length.
+    pub fn normalize(&self) -> Self {
+        let n = self.norm();
+        assert!(n > 0.0, "cannot normalize a zero-length vector");
+        self.scale(1.0 / n)
+    }
+
+    /// Try to normalize to a unit vector, returning `None` if the vector has
+    /// zero length.
+    pub fn try_normalize(&self) -> Option<Self> {
+        let n = self.norm();
+        if n > 0.0 {
+            Some(self.scale(1.0 / n))
+        } else {
+            None
+        }
+    }
+}
+
+/// 3D coordinates for all heavy atoms in a molecule.
+///
+/// Indexed by atom insertion order (`AtomIdx.0` as `usize`). A dumb data
+/// holder only -- no embedding, minimization, or alignment logic (those stay
+/// in `chematic-3d`).
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct Coords3D {
+    pub points: Vec<Point3>,
+}
+
+impl Coords3D {
+    /// Create zeroed coordinates for `n` atoms.
+    pub fn new_zeroed(n: usize) -> Self {
+        Self {
+            points: vec![Point3::zero(); n],
+        }
+    }
+
+    /// Get the coordinate of atom `idx`.
+    ///
+    /// # Panics
+    /// Panics if `idx` is out of range (matching `chematic_3d::coords::Coords3D`'s
+    /// existing indexing convention).
+    pub fn get(&self, idx: AtomIdx) -> Point3 {
+        self.points[idx.0 as usize]
+    }
+
+    /// Set the coordinate of atom `idx`.
+    ///
+    /// # Panics
+    /// Panics if `idx` is out of range.
+    pub fn set(&mut self, idx: AtomIdx, p: Point3) {
+        self.points[idx.0 as usize] = p;
+    }
+
+    /// Number of atom coordinate slots.
+    pub fn atom_count(&self) -> usize {
+        self.points.len()
+    }
+
+    /// `true` when every point has all-finite components.
+    pub fn is_finite(&self) -> bool {
+        self.points.iter().all(Point3::is_finite)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn new_zeroed_has_n_points_at_origin() {
+        let c = Coords3D::new_zeroed(3);
+        assert_eq!(c.atom_count(), 3);
+        for i in 0..3 {
+            assert_eq!(c.get(AtomIdx(i)), Point3::zero());
+        }
+    }
+
+    #[test]
+    fn get_set_roundtrip() {
+        let mut c = Coords3D::new_zeroed(2);
+        c.set(AtomIdx(1), Point3::new(1.0, 2.0, 3.0));
+        assert_eq!(c.get(AtomIdx(1)), Point3::new(1.0, 2.0, 3.0));
+        assert_eq!(c.get(AtomIdx(0)), Point3::zero());
+    }
+
+    #[test]
+    fn is_finite_true_for_normal_coords() {
+        let mut c = Coords3D::new_zeroed(1);
+        c.set(AtomIdx(0), Point3::new(1.0, -2.5, 0.0));
+        assert!(c.is_finite());
+    }
+
+    #[test]
+    fn is_finite_false_for_nan() {
+        let mut c = Coords3D::new_zeroed(1);
+        c.set(AtomIdx(0), Point3::new(f64::NAN, 0.0, 0.0));
+        assert!(!c.is_finite());
+    }
+
+    #[test]
+    fn is_finite_false_for_infinite() {
+        let mut c = Coords3D::new_zeroed(1);
+        c.set(AtomIdx(0), Point3::new(0.0, f64::INFINITY, 0.0));
+        assert!(!c.is_finite());
+    }
+
+    #[test]
+    fn point3_basic_vector_ops() {
+        let a = Point3::new(1.0, 0.0, 0.0);
+        let b = Point3::new(0.0, 1.0, 0.0);
+        assert_eq!(a.cross(&b), Point3::new(0.0, 0.0, 1.0));
+        assert_eq!(a.dot(&b), 0.0);
+        assert_eq!(a.distance(&b), std::f64::consts::SQRT_2);
+        assert_eq!(a.add(&b), Point3::new(1.0, 1.0, 0.0));
+    }
+}
--- a/crates/chematic-core/src/coords3d.rs
+++ b/crates/chematic-core/src/coords3d.rs
@@ -23,10 +23,18 @@
 //! touched by this PR (read-only reference); the two types are reconciled at
 //! Wave 1->2 integration time, not here.
 //!
-//! This module adds exactly one thing `chematic_3d::coords` doesn't have:
-//! `is_finite()` on both types, since `chematic-mol`'s readers need a way to
-//! reject NaN/Inf coordinates as a typed parse error rather than silently
-//! constructing a poisoned conformer.
+//! This module adds two things `chematic_3d::coords` doesn't have:
+//! - `is_finite()` on both types. **Not currently called by this PR's own
+//!   reader code** -- `chematic-mol`'s V2000/V3000 z-coordinate parsers
+//!   validate NaN/Inf with a direct `f64::is_finite()` check on the raw
+//!   parsed value, *before* a `Point3`/`Coords3D` is ever constructed (see
+//!   `mol2000.rs`/`mol3000.rs`), so these methods have zero non-test callers
+//!   today. Provided for future consumers (e.g. Wave 2 work that may
+//!   construct a `Coords3D` from elsewhere and want to validate it after the
+//!   fact) rather than because this PR's own code needs them.
+//! - `Default`/`PartialEq` derives on `Coords3D` (`chematic_3d::coords::Coords3D`
+//!   derives only `Debug, Clone`) -- needed for this crate's own tests and
+//!   generally harmless additions for a plain data holder.
 
 use crate::molecule::AtomIdx;
 
--- a/crates/chematic-core/src/lib.rs
+++ b/crates/chematic-core/src/lib.rs
@@ -16,6 +16,7 @@
 
 pub mod atom;
 pub mod bond;
+pub mod coords3d;
 pub mod element;
 pub mod kekulization;
 pub mod molecule;
@@ -25,6 +26,7 @@ pub mod valence;
 // Re-export the most commonly used types at crate root.
 pub use atom::{Atom, Chirality, CipCode};
 pub use bond::{BondEntry, BondOrder};
+pub use coords3d::{Coords3D, Point3};
 pub use element::Element;
 pub use kekulization::{KekuleError, KekuleResult, apply_kekule, kekulize};
 pub use molecule::{AtomIdx, BondIdx, MolError, Molecule, MoleculeBuilder, STEREO_H_SENTINEL};
--- a/crates/chematic-mol/src/lib.rs
+++ b/crates/chematic-mol/src/lib.rs
@@ -43,14 +43,20 @@ pub use gaussian::{GaussianError, GaussianLogResult, parse_gaussian_log, parse_g
 pub use ket::{KetError, parse_ket, parse_ket_3d, write_ket, write_ket_3d};
 pub use mol2_tripos::{Mol2Error, parse_mol2, write_mol2};
 pub use mol2000::{
-    MolMetadata, MolReadReport, parse_mol, parse_mol_with_coords, parse_sdf_with_coords,
-    read_mol_with_diagnostics, read_sdf_with_diagnostics, write_mol, write_mol_with_coords,
-    write_sdf, write_sdf_record, write_sdf_record_v3000, write_sdf_with_charges,
+    CoordinateDimension, GeometryRank, MolMetadata, MolReadReport, Stereo3DDiagnostic, parse_mol,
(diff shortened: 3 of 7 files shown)


kent-tokyo:
Addressed all 6 points from independent verification, pushed as `9cb6192`:

1. Fixed the "244/244 pre-existing" mislabel in the PR body — corrected to 238 pre-existing (99 core + 139 mol lib tests), independently re-confirmed via a `comm`-diff of sorted `--list` test names (0 removed/renamed, 6 added), + 33 new (6 lib + 27 integration) = 271 total. No "279"/"312" figures were ever cited in the PR body, so nothing to correct there.
2. Corrected the `is_finite()` doc comment in `coords3d.rs` — confirmed via grep that `mol2000.rs:504`/`mol3000.rs:334` call raw `f64::is_finite()` directly, not the new methods (zero non-test callers today). Also documented the `Default`/`PartialEq` derives added beyond `chematic_3d::coords::Coords3D`'s shape.
3. Added a "Behavior change (semver-relevant)" section to the PR body with a suggested `CHANGELOG.md` line (not editing that file myself, per file ownership) covering the new `Err(InvalidAtomLine)` on garbled/non-finite z.
4. Noted as ratified, no action needed.
5. Added root re-exports for `CoordinateDimension`/`GeometryRank`/`Stereo3DDiagnostic`/the new writer fns/`ConformerEnsemble`/`read_sdf_conformer_ensembles` to `crates/chematic-mol/src/lib.rs`, matching the existing `MolReadReport`/`SdfRecord` pattern.
6. Filed https://github.com/kent-tokyo/chematic/issues/171 for the pre-existing SDF blank-name-line bug, with root-cause line citations (`mol2000.rs:715`, `sdf.rs:40`/`163`) and the historical partial workaround (`chematic-py/src/formats.rs:248`, since v0.4.10) — not fixed here, per scope discipline.

Re-verified after these changes: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings` (full workspace, incl. chematic-py), `cargo test -p chematic-mol -p chematic-core --lib` (244/244), `cargo test -p chematic-mol --test conformer_3d_io` (27/27), `python3 scripts/check_publish_graph.py` — all still pass. Working tree clean, pushed.

Issue #161, opened 2026-07-25
verified dedup: use accurate CIP preflight to reduce conservative VerificationUnavailable


kent-tokyo:
## Background

PR #156's `has_unresolved_specified_tetrahedral_stereo` guard fails `identity_verify` closed to `VerificationUnavailable` whenever the **legacy** CIP engine (`chematic_chem::assign_cip`, `~96.3%` oracle agreement) can't rank a specified tetrahedral stereocentre — necessary to close a real false-`VerifiedDuplicate` (corpus rows 4663/4664).

A full-corpus audit of every molecule this guard newly fails closed (`validation/results/dedup_stereo_guard_corpus_audit.jsonl` / `_summary.json`, 12 molecules / 28 atoms) found:

- **10 molecules / 26 atoms**: legacy CIP fails, but the **accurate** engine (`chematic_chem::assign_cip_with_mode(mol, CipMode::Accurate)`, `chematic-cip` crate) resolves them — and its resolved label matches an independent RDKit 2026.03.3 oracle (`rdCIPLabeler`) exactly in every case (plain R/S and Rule-5 pseudoasymmetric r/s alike).
- **2 molecules / 2 atoms** (tricyclic-cage bridgeheads, idx 443/590): unresolved by *both* chematic engines, and RDKit's own modern CIP labeler agrees no label exists (`Tet_CCW`) — a genuine, shared CIP-ranking limit, not fixable by an accurate-CIP preflight.

## Proposal

Add an opt-in preflight in `chematic-inchi::dedup` that, before falling back to `VerificationUnavailable` on a legacy-CIP-unresolved centre, re-checks the atom via `CipMode::Accurate`. If the accurate engine resolves it (not `Tied`/`BudgetExceeded`), treat the centre as ranked for comparison purposes instead of failing closed — recovering verified-comparison capability for the ~10/12 (26/28 atom) majority of cases audited here, while still failing closed for genuine ties (case B, 2/12 molecules) where even the accurate engine can't produce a label.

## Scope notes (from the #156 audit)

- This does **not** mean switching native InChI generation itself to the accurate engine wholesale — that would change `standard_inchi`'s output for every caller, not just `dedup`, and needs its own proposal given the accurate engine's budget/tied-result semantics would need threading through the whole native-conversion Stereo0D pipeline.
[...]

PR #170, opened 2026-07-26
feat(inchi): accurate-CIP dedup preflight (#161) + exact graph identity API

kent-tokyo:
## Summary

Two deliverables from the 3D Breakthrough Program's Agent B brief, both scoped to `crates/chematic-inchi/src/dedup.rs` + `crates/chematic-inchi/src/native/**` + new `crates/chematic-inchi/tests/**` files only (no other file touched):

1. **Closes issue #161** — an accurate-CIP preflight for the verified-dedup fail-closed guard shipped in PR #156.
2. **New, standalone "exact graph identity" API** (`GraphIdentityMode` / `GraphRelation` / `IdentityEvidence` / `compare_graph_identity`) for conformer/ensemble grouping, MOL/SDF round-tripping, and dedup to share (Coordinator wires this into Agent A's SDF-ensemble supplier at Wave 2 integration).

- Base SHA: `9f9f459` (`main`, v0.7.0)
- Head SHA: `d0c9fbd`
- Commits (3, each independently reviewable):
  - `c2db0ea` `feat(inchi): add accurate-CIP dedup preflight (issue #161)`
  - `768e4cb` `feat(identity): add exact graph relation API`
  - `d0c9fbd` `test(inchi): directly assert the rank-collision guard's failure reason` — added after a pre-PR self-review pass; see "Known limitation" in §1 below for what it closes
- Files touched (exactly, nothing else):
  - `crates/chematic-inchi/src/dedup.rs`
  - `crates/chematic-inchi/src/native/convert.rs`
  - `crates/chematic-inchi/src/native/mod.rs`
  - `crates/chematic-inchi/tests/accurate_cip_preflight.rs` (new)
  - `crates/chematic-inchi/tests/graph_identity.rs` (new)

**Deviation from the coordinator's suggested 3-commit shape**, flagged as instructed: the coordinator suggested a 3rd commit `test(validation): add identity differential corpus`. My file-ownership scope for this PR is exactly `dedup.rs` + `native/**` + new files under `tests/**` — I have no permission to add anything under `validation/` (that directory isn't in my allow-list, and isn't the crate I own). I folded the RDKit-oracle-backed fixture/regression tests into the same commit as the feature they validate instead (`tests/accurate_cip_preflight.rs` → commit 1, `tests/graph_identity.rs` → commit 2). The 3rd commit that does exist (`d0c9fbd`) is unrelated to that suggestion — it's a small review-driven fix (see below), not the validation-corpus commit the coordinator meant. Each commit is still independently buildable/lintable/testable (verified below) and maps to one conceptual change.

---

### Root cause (why a naive guard relaxation is unsafe)

PR #156's `has_unresolved_specified_tetrahedral_stereo` guard fails `identity_verify` closed whenever the **legacy** CIP engine (`chematic_chem::assign_cip`) can't rank a specified tetrahedral stereocentre's substituents — because `crate::native::convert::mol_to_inchi_atoms` depends on that *exact same* ranking function (`tetrahedral_stereo_neighbors`) to build the InChI `Stereo0D` descriptor. When ranking fails, native InChI generation doesn't just decline gracefully — empirically (audit data below), it emits either a literal `?` (undefined parity) if *some* centre in the molecule succeeded, or omits the whole `/t` layer if *none* did:

```
idx=196 chematic:  .../t19-,20+,26?          (2/3 centres resolved, 1 shows '?')
idx=443 chematic:  ...no /t layer at all      (0/4 centres resolved)
```

Issue #161 proposed: "re-check a legacy-unresolved centre via `CipMode::Accurate`; if it resolves, treat the centre as ranked instead of failing closed." Taken literally (just relax the guard, then call `standard_inchi()` as before), this is **unsafe**: `standard_inchi()`'s own Stereo0D construction still depends on the *legacy* ranking regardless of what the accurate engine says, so it would still emit the same `?`/missing-`/t` output for that centre — reopening the exact 4663/4664 false-`VerifiedDuplicate` PR #156 fixed. Issue #161 itself anticipates this ("this does NOT mean switching native InChI generation to the accurate engine wholesale").

### Design: fold the accurate code into the comparison key, not into `standard_inchi()`

New, **additive-only** entry points — the existing `compare` / `compare_molecules` / `deduplicate_verified` / `identity_verify` / `verified_identity_key` are untouched (verified byte-for-byte via `preflight_is_opt_in_existing_paths_unchanged`):

```rust
pub fn compare_with_accurate_cip_preflight(key_a, mol_a, key_b, mol_b, policy) -> DedupRelation
pub fn compare_molecules_with_accurate_cip_preflight(mol_a, mol_b, policy) -> DedupRelation
pub fn deduplicate_verified_with_accurate_cip_preflight(mols, policy) -> VerifiedDedupReport
```

For each specified stereocentre the legacy engine can't rank, `accurate_stereo_supplement` re-checks via `chematic_chem::assign_cip_with_mode(mol, CipMode::Accurate)`. If the accurate engine resolves it (not `Tied`/`BudgetExceeded`/error), its code is folded into the **comparison key** — never into the generated InChI string, and never into the `InChIKey` C-library call (`verified_identity_key_with_supplement` reuses the existing, unmodified `verified_identity_key` for the base key, then appends the supplement afterward, so `StandardInchiKey`'s FFI call always still receives a clean, real InChI string).

Cross-molecule correspondence for the supplement uses `chematic_smiles::morgan_ranks` (a purely constitutional, stereo-blind isomorphism invariant, already used/tested elsewhere in this codebase) rather than InChI's own canonical numbering (which would require parsing `AuxInfo`, out of scope) or a full atom-mapping search. **Fails closed** (keeps `VerificationUnavailable`) if the accurate engine ties/budgets-out/errors on *any* flagged atom, or — a case the original audit didn't anticipate — if **two or more flagged atoms in the same molecule land on the same morgan-rank value** (their cross-molecule correspondence would be ambiguous; see limitation below).

### Before/after, re-derived fresh (not trusted from the old audit)

PR #156's own audit (`validation/results/dedup_stereo_guard_corpus_audit.jsonl`) is the fixture starting point issue #161 names — but I did **not** take its 10/12 recovery number on faith. Re-running the exact same 12 audited molecules (pinned literal SMILES, not re-read from the mutable corpus file) against current HEAD found the classification has **partially drifted since the audit ran**, for two independent, unrelated reasons:

| Reason | Molecules affected |
|---|---|
| Upstream `chematic-chem`/`chematic-cip` CIP-ranking change since the audit ran — legacy now resolves this molecule's centres, guard doesn't fire at all any more | 1 (audit idx=4412) |
| A 3-fold-symmetric substituent (adamantane-cage-like) whose legacy-unresolved centres all land on the **same morgan rank** — the accurate engine resolves each individually, but my correspondence guard (above) declines to pair them | 3 (audit idx=1609, 1643, 4178) |

I then independently re-ran the **full 5,000-molecule corpus** (`~/Downloads/SMILES.csv`, current local snapshot — see "Dataset provenance" below for why this differs from the audit's own numbering) through `deduplicate_verified` vs. `deduplicate_verified_with_accurate_cip_preflight`:

| Metric | Plain (`compare`/`deduplicate_verified`, unchanged) | With accurate-CIP preflight |
|---|---|---|
| `verification_unavailable` (of 5,000) | **15** | **6** |
| Recovered (→ `VerifiedDuplicate`) | — | **9** |
| Newly made unavailable | — | **0** (safety invariant, confirmed empirically over the whole corpus) |

The 6 remaining `VerificationUnavailable` molecules, fully explained (not left as an unattributed residual):

| Count | Reason |
|---|---|
| 1 | Pre-existing native-InChI `KekulizationFailed` (unrelated to this guard entirely — a fused/bridgehead heteroaromatic ring chematic can't Kekulize; same molecule the original audit's own scope note excluded at idx=2149) |
| 2 | Genuine CIP tie in **both** chematic engines (tricyclic-cage bridgeheads) |
| 3 | Correctly resolved by the accurate engine, but blocked by the tied-morgan-rank correspondence guard (see limitation below) |

### Independent RDKit 2026.03.3 cross-check (own isolated venv, never the shared `.venv`)

- Isolated venv: `/private/tmp/claude-501/-Users-k-tanabe-Documents-Documents-oss-rust-chematic/9934aa14-7af8-40c0-850a-54a336991cb0/scratchpad/dedup_b_venv`, built from `/opt/homebrew/bin/python3.13` (a clean system interpreter, **not** the repo's shared `.venv` at `/Users/k_tanabe/Documents/Documents/oss_rust/chematic/.venv`, and not any other agent's worktree venv). `pip install "rdkit==2026.03.3"` → confirmed `Chem.rdBase.rdkitVersion == "2026.03.3"`, matching the version pin in this task's brief.
- All **9 recovered** molecules' accurate-CIP codes (15 atoms total) cross-checked against `rdCIPLabeler.AssignCIPLabels` + `Chem.FindMolChiralCenters(includeUnassigned=True, useLegacyImplementation=False)`: **15/15 agree exactly** (R/S/r/s, full denominators, no pooling of a subset as if it were the whole set).
- The 2 genuine-tie molecules: RDKit's own modern CIP labeler independently confirms `Tet_CCW`/no label at the exact same bridgehead atom — a real, shared CIP-ranking limit in both engines, not a chematic defect.
- The 3 tied-morgan-rank-blocked molecules: RDKit assigns `s` to all 9 atoms (3 molecules × 3 symmetric centres), matching chematic's accurate engine exactly — confirming these ARE individually correct, and that the recall cost is purely from my conservative correspondence guard, not from an underlying CIP error.

### Known limitation: tied-morgan-rank correspondence (real, quantified, not swept under the rug)

`accurate_stereo_supplement` fails closed whenever two-or-more legacy-unresolved centres in the *same* molecule share a `morgan_ranks` value (constitutionally symmetric positions, e.g. a 3-fold-symmetric cage substituent). This is **safe by construction under the guard**: a spurious match would require two flagged atoms with swapped codes colliding on one rank key, and the `seen_ranks.insert` check returns `Err(AmbiguousStereoRankCorrespondence)` on exactly that collision before any code is pushed to the supplement — not asserted as a proof, but directly observed: commit `d0c9fbd`'s `accurate_stereo_supplement_blocks_via_rank_collision_not_some_other_path` calls the private function directly on idx=1609 and asserts the *exact* `Err` variant, ruling out `AccurateCipTied`/`AccurateCipBudgetExceeded`/`AccurateCipEngineError` as the real cause (a prior version of this PR only checked the outcome via `compare_molecules_with_accurate_cip_preflight`, which couldn't distinguish "blocked by the rank guard" from "blocked by some other path" — caught in self-review before opening this PR). This costs real recall — 3/12 audited molecules, 3/15 corpus molecules stay `VerificationUnavailable` even though every individual centre is correctly resolved and RDKit-confirmed. A rank-tie-aware correspondence mechanism that additionally proves automorphism-consistency (not just "same label multiset") could recover these safely, but that's a nontrivial, separate piece of work — flagged as a follow-up, not attempted here given the risk of getting the automorphism argument subtly wrong under time pressure.

### Dataset provenance note (a real, orthogonal finding, not excuse-making)

`~/Downloads/SMILES.csv` (5,000 lines, no header) is the same path PR #156's own audit used, but **not byte-identical to whatever produced that audit's JSONL** — the file on this machine has drifted (most of the audit's `corpus_index`-N molecules now sit at position N+1 in the current file; a handful of positions near the audit's idx≈4412 contain genuinely different molecules, not just a shift). This is exactly the kind of "don't trust old numbers/artifacts on faith" situation this project has hit before (see the Wave 0 master plan's own re-verification of `etkdg_vs_rdkit_gap.py`). My headline 5,000-molecule numbers above are from **this PR's own fresh run against the file as it exists now**, not backfilled from the old audit; my 12-molecule fixture uses **pinned literal SMILES** (not re-read from the mutable CSV), so it is reproducible regardless of future corpus-file drift.

---

### Shape

```rust
pub struct GraphIdentityMode {
    pub graph_strictness: GraphStrictness,  // RawGraphExact | ChemicalGraphExact
    pub atom_maps: AtomMapPolicy,           // Include | Ignore
}
pub enum GraphRelation { Identical, Distinct }
pub struct IdentityEvidence {
    pub graph_relation: Option<GraphRelation>,
    pub standard_inchi: Option<String>,
    pub inchikey: Option<String>,
    pub diagnostics: Vec<IdentityDiagnostic>,
}
pub fn compare_graph_identity(mol_a: &Molecule, mol_b: &Molecule, mode: GraphIdentityMode) -> IdentityEvidence
```

### Two orthogonal axes, not one flat enum (per the coordinator's design-review refinement)

- **`GraphStrictness`** — how strictly bond order/aromaticity spelling is compared:
  - `RawGraphExact`: literal bond-order/`aromatic`-flag equality. Aromatic-order-spelled benzene ≠ an explicit Kekulé respelling of the same ring under this mode (by design — this is the strictest, most literal notion of "same graph as written").
  - `ChemicalGraphExact`: any input with `Aromatic`-order bonds is Kekulized first via `chematic_core::kekulize`/`apply_kekule` — the same algorithm this crate's own native InChI conversion already uses (not a new one written for this PR), rather than assuming its determinism generalizes; only directly confirmed for the fixtures this PR actually exercises (e.g. `graph_identity_raw_graph_exact_distinguishes_aromatic_spelling`'s benzene case), not proven in general — and the `aromatic` flag itself is excluded from atom comparison (Kekulization never clears it, so comparing it post-Kekulization would silently reintroduce the exact spelling-sensitivity this mode exists to remove). **Never implemented via canonical-SMILES string equality** — this module's standing rule (canonical SMILES is a fast pre-filter elsewhere in this file, never the proof) applies here too.
- **`AtomMapPolicy`** — `Include`/`Ignore` reaction atom-mapping numbers (`Atom::atom_map`), fully independent of the strictness axis. All 4 combinations are independently selectable and tested (`graph_identity_atom_map_policy_is_respected`, `atom_map_axis_is_orthogonal_to_strictness_axis`).

### How identity is actually verified (never canonical SMILES)

1. **Cheap, atom-order-independent composition pre-check**: fragment sizes, element/isotope/charge/aromaticity multiset, bond-order multiset. A mismatch here is conclusive `Distinct` regardless of atom ordering — a real graph invariant, not a string comparison.
2. **Direct input-index correspondence + full bond-set check**: if atom `i` in molecule A is assumed to correspond to atom `i` in molecule B, does *every* bond line up (not just counts)? If yes, every remaining atom/bond attribute is verified directly at each position — conclusive either way. If the bond sets don't line up under this direct correspondence, the result is **`None` (inconclusive)**, never a guess.
3. **Tetrahedral parity**: a from-scratch, self-contained *permutation-parity* comparison (`canonical_tetrahedral_parity`) over `Molecule::stereo_neighbor_order`, expressed in a common ascending-index reference frame — no CIP engine involved at all (this is deliberately a different, lower-level tool than the accurate-CIP preflight above; graph identity needs pure combinatorial parity, not chemical priority ranking).
4. **E/Z**: directed `Up`/`Down` bond-marker comparison (`directed_bond_marker`), mirroring `crate::native::convert::mol_to_inchi_atoms`'s own `is_up` sign convention exactly (not a new one), checked over every adjacent atom pair — no reference-substituent-choice ambiguity, since every pair is checked directly rather than picking one.
5. `standard_inchi`/`inchikey` in `IdentityEvidence` are **independent corroborating evidence only** (each molecule's own full-stereo native InChI, populated only when both succeed and literally agree) — never the basis for `graph_relation` itself. Native InChI's own tautomer/mobile-H normalization is a different, weaker question than literal graph identity (same distinction this module's `IdentityPolicy` docs already make).

### Scope boundaries (documented, not silently cut)

- **Atom ordering**: the structural comparator only recognizes identity when the two molecules share the same atom indexing (index `i` ↔ index `i`) — not a general graph-isomorphism search over arbitrary permutations. A real isomorphism search needs either `chematic-smarts`'s VF2 (not on `chematic-inchi`'s dependency graph — adding it means editing `Cargo.toml`, outside this PR's file-ownership scope) or a from-scratch bounded backtracking matcher with its own correctness-testing burden I judged too risky to build and verify within this PR's budget. The task's own stated real-world consumer (conformer/ensemble grouping over records sharing one connection table) already satisfies same-indexing by construction, so this boundary doesn't block that use case. When ordering doesn't line up, `graph_relation` is `None` with `IdentityDiagnostic::AtomOrderCorrespondenceNotEstablished` — verified never to silently guess `Identical` or `Distinct` (`graph_identity_reordered_atoms_is_inconclusive_not_a_guess`, using `random_smiles` respellings).
- **`ChemicalGraphExact` and alternative resonance structures**: Kekulizing normalizes "aromatic-flagged" vs. "one particular already-Kekulized" spelling, but two *independently-written*, already-Kekulized alternative resonance structures of the same aromatic system (e.g. two different valid Kekulé structures of naphthalene) are not recognized as equivalent — that needs Hückel aromaticity perception (`chematic_perception::apply_aromaticity`), which is also not on `chematic-inchi`'s dependency graph. Never a false `Identical` either way (`ChemicalGraphKekulizationFailed` fails closed to inconclusive if Kekulization itself errors).
- **Enhanced stereochemistry**: the task brief asked for "enhanced stereo (if representable)" to be compared. Grepped `chematic-core`/`chematic-perception` before writing this module: neither has any AND/OR/ABS-style relative-stereo-group representation beyond plain per-atom `Chirality` today. Documented as a real scope boundary; no new enhanced-stereo model was invented as a side effect of this task.

---

## Commands run

```bash
# Isolated venv (own scratchpad, never the shared repo .venv):
/opt/homebrew/bin/python3.13 -m venv <scratchpad>/dedup_b_venv
<scratchpad>/dedup_b_venv/bin/pip install "rdkit==2026.03.3"

cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings
cargo build -p chematic-inchi [--features native-inchi]
cargo test  -p chematic-inchi [--features native-inchi] --lib --quiet
cargo test  -p chematic-inchi --features native-inchi --test dedup_fixtures --quiet
cargo test  -p chematic-inchi --features native-inchi --test accurate_cip_preflight --quiet
cargo test  -p chematic-inchi [--features native-inchi] --test graph_identity --quiet
cargo test  --workspace --tests --quiet --exclude chematic-py     # rest of the workspace, unaffected
cargo deny --all-features check
python3 scripts/check_publish_graph.py
```

`chematic-py` cannot be built/tested via plain `cargo build`/`cargo test` in this environment at all (fails at the link stage — missing `libpython` symbols) regardless of this PR's changes; `CLAUDE.md` itself documents that it requires `maturin develop`, which this task explicitly forbids running against the shared `.venv`. `cargo clippy` (type-check only, no final link) succeeds for `chematic-py` and every other crate workspace-wide. I did not build/install/touch the shared `.venv` at any point.

## Confirmation: shared `.venv` never touched

All RDKit oracle work ran through `/private/tmp/claude-501/.../scratchpad/dedup_b_venv`, built from a clean system Python (`/opt/homebrew/bin/python3.13`), never `/Users/k_tanabe/Documents/Documents/oss_rust/chematic/.venv` and never `maturin develop`/`pip install -e`/`pytest` against chematic's own Python bindings.

## Known limitations / follow-ups (not filed as GitHub issues — described here for the Coordinator to triage)

1. Tied-morgan-rank correspondence guard (3/12 audited, 3/15 fresh-corpus molecules stay conservatively unavailable) — see above.
2. `GraphIdentityMode`'s atom-order-identity-only scope — a real isomorphism search is a natural, separate follow-up once file ownership allows a `chematic-smarts` dependency edge (or a from-scratch matcher is independently designed/reviewed).
3. `ChemicalGraphExact` doesn't recognize alternative Kekulé resonance structures without Hückel perception access.
4. Enhanced stereochemistry comparison: blocked entirely on `chematic-core`/`chematic-perception` not having a representation yet.

## Merge blockers / open questions for the Coordinator

- Please independently review the tied-morgan-rank recall trade-off (limitation 1) — I judged "fail closed, ship less recall" as correct given this codebase's own stated fail-closed philosophy, but it's a judgment call worth a second opinion.
- `GraphIdentityMode`/`IdentityEvidence` is new, unintegrated API — Coordinator wires it into Agent A's SDF-ensemble supplier at Wave 2 per the master plan; no code in this PR depends on Agent A's work or vice versa.
- Not self-merging, per instructions.
- Side note for the Coordinator, not specific to this PR: the shared scratchpad directory (`/private/tmp/claude-501/.../scratchpad/`) is program-wide, not per-agent — I found a pre-existing `pr_body.md` there from a different agent's work and used a distinctly-named `pr_body_agent_b.md` instead to avoid clobbering it. Worth a heads-up to the other agents/coordinator if nobody's flagged this already.

### DIFF
--- a/crates/chematic-inchi/src/dedup.rs
+++ b/crates/chematic-inchi/src/dedup.rs
@@ -127,7 +127,7 @@
 
 #[cfg(feature = "native-inchi")]
 use chematic_core::AtomIdx;
-use chematic_core::Molecule;
+use chematic_core::{CipCode, Molecule};
 use chematic_smiles::canonical_smiles;
 use std::collections::HashMap;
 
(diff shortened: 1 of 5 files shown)


kent-tokyo:
Superseded by #178 and #181.

- #178 contains the corrected accurate-CIP preflight (issue #161), with the idx=4412 factual error fixed and the previously-missing white-box rank-collision regression test ported over.
- #181 contains the renamed indexed graph relation API (`compare_indexed_graph_relation` / `IndexedGraphRelationMode`), split out from this PR's original "exact graph identity" work.

Both replacements are independently verified and merged. Closing this one as superseded with no remaining unported content.

PR #198, opened 2026-07-29
fix(mol): stop stripping legal blank MOL name lines in SDF splitting

### LINKED ISSUE #171: SDF reader: leading-blank-line-skip strips a legitimately-blank MOL name line, breaking unnamed-molecule SDF (3 entry points)
## Summary

`chematic-mol`'s string-based SDF `$$$$`-block-splitters unconditionally strip *leading* blank lines before extracting each record — including a legitimately blank **first line** of a record (the MOL name line), which is the literal default output shape of RDKit's own `Chem.MolToMolBlock()` for an unnamed molecule. This shifts every subsequent line of that record up by one and causes a hard parse error (`InvalidCountLine { detail: "missing V2000 version tag" }` or similar) on any SDF containing an unnamed molecule.

## Root cause

Three call sites share the same "skip leading blank lines" preprocessing loop, applied unconditionally at the top of each `next()`/parse iteration — not just between records, but also before the very first record and before every subsequent one:

- `crates/chematic-mol/src/mol2000.rs:715` — `read_sdf_with_diagnostics` (and therefore `parse_sdf_with_coords`)
- `crates/chematic-mol/src/sdf.rs:40` — `SdfReader::next` (and therefore `parse_sdf`)
- `crates/chematic-mol/src/sdf.rs:163` — `SdfRecordReader::next`

```rust
// Skip leading blank lines.
while let Some(rest) = remaining
    .strip_prefix("\r\n")
    .or_else(|| remaining.strip_prefix('\n'))
{
    remaining = rest;
}
```

This is meant to defensively skip stray blank lines *between* `$$$$`-delimited records, but it also fires on the very first line of whatever `remaining` currently points at — which, for a record whose own name (header line 1) is blank, eats that legitimate blank line, shifting the rest of that record's 3-line header out of alignment with the parser's fixed line-number expectations.

`SdfFileReader` (the `BufRead`-based streaming reader, `crates/chematic-mol/src/sdf.rs`) does **not** have this bug — it accumulates lines verbatim into `block` without any leading-blank-line skip, so a blank name line survives intact there.

## Repro

```rust
let block = "\n     RDKit          3D\n\n  1  0  0  0  0  0  0  0  0  0  0 V2000\n    0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0\nM  END\n";
let sdf = format!("{block}$$$$\n");
assert!(chematic_mol::parse_sdf(&sdf).is_err()); // fails with InvalidCountLine, not a real error in the input
```

Any SDF produced by `Chem.MolToMolBlock(unnamed_mol)` (or any writer that emits a blank name line, which is common and spec-legal) trips this on every one of the 3 affected entry points.

## Known partial workaround (pre-existing, since v0.4.10)

`crates/chematic-py/src/formats.rs`'s `parse_sdf_with_coords` (the PyO3-exposed batch function) already works around this at the Python-binding layer by implementing its own independent `$$$$`-splitting loop that does **not** skip leading blank lines, specifically to avoid calling into the buggy Rust `parse_sdf_with_coords`/`read_sdf_with_diagnostics` — see the comment at `crates/chematic-py/src/formats.rs:248`: `// This avoids the Rust parse_sdf_with_coords leading-blank-line stripping issue.` This workaround only covers that one Python-exposed function; `SdfReader`, `SdfRecordReader`, `read_sdf_with_diagnostics`, and every other Rust-side or Python-side caller of the 3 affected entry points above are still exposed to the bug. The underlying Rust splitter bug itself was never fixed at the root.

## Found via

Found incidentally in PR #168 (3D Breakthrough Program, Agent A) while building a real, RDKit-generated multi-conformer SDF test fixture — RDKit's own `Chem.MolToMolBlock()` output for an unnamed molecule starts with a blank name line, which tripped this on the first attempt. Worked around in that PR's own test (by giving the fixture molecules explicit names) rather than fixed there, since the bug is pre-existing, general SDF-delimiter-handling, and unrelated to 3D coordinates specifically.

## Proposed fix (not attempted here — scope discipline)

The leading-blank-line skip should only run *between* records (i.e. after consuming a `$$$$` delimiter and before searching for the next one), never as a blanket first step of every iteration regardless of position. A `first_iteration` flag, or restructuring the loop so the skip only happens post-delimiter, would fix all 3 sites without changing behavior for any currently-passing fixture (all of which use non-blank names).

kent-tokyo:
## Summary

Fixes #171. `read_sdf_with_diagnostics`, `SdfReader`, and `SdfRecordReader` each
unconditionally skipped leading blank lines before searching for the next
`$$$$` delimiter. This was meant to defensively absorb stray inter-record
padding, but it also ate a record's own **legitimately blank first line**
(the MOL name line) — which is exactly what RDKit's own
`Chem.MolToMolBlock()` emits for an unnamed molecule. Any SDF containing an
unnamed molecule (first, middle, or last record) failed to parse on all
three affected entry points.

## Root cause + fix

The pre-skip was redundant: a genuinely empty gap between/after `$$$$`
delimiters is already caught downstream by the existing
`block.trim().is_empty() => skip` check in all three functions. Removing the
pre-skip loop fixes the blank-name-line bug without adding any new
special-casing — verified this doesn't change behavior for any existing
passing fixture (none of them exercise a stray-blank-line-between-records
scenario).

`SdfFileReader` (the streaming `BufRead` reader) never had this bug — it
accumulates lines verbatim without any leading-blank-line skip.

## Testing

- New regression tests cover first/middle/last record position, LF and CRLF
  line endings, across all 3 affected entry points
  (`read_sdf_with_diagnostics`, `parse_sdf_with_coords`, `SdfReader`/`parse_sdf`,
  `SdfRecordReader`).
- Fixture is a **live RDKit 2026.03.3** `Chem.MolToMolBlock()` output for an
  unnamed molecule (`AllChem.Compute2DCoords(Chem.MolFromSmiles("CC"))`), not
  a hand-typed approximation.
- Added explicit malformed-input regression tests (bad counts line still
  errors) to confirm existing error recovery is unaffected.
- `cargo test -p chematic-mol --lib`: 152/152 pass (up from 141 pre-change +
  11 new).
- `bash scripts/check.sh`: full pass (fmt, clippy `-D warnings`, workspace
  tests, cargo-deny, publish-graph/version checks).

## Test plan

- [x] Blank first line preserved as legal MOL name line
- [x] Extra inter-record blank lines still skipped (existing behavior)
- [x] LF and CRLF
- [x] First, middle, and last record
- [x] `read_sdf_with_diagnostics`, `SdfReader`, `SdfRecordReader` all covered
- [x] RDKit-generated blank-name MOL/SDF fixture as regression test
- [x] Existing malformed-input error recovery unaffected
- [x] `bash scripts/check.sh` passes


kent-tokyo:
Added a second commit (14539cf): the Python binding's \`parse_sdf_with_coords\` had its own hand-rolled \`\$\$\$\$\`-splitting loop that existed *only* to work around the Rust core bug this PR fixes (see the removed comment: "avoids the Rust parse_sdf_with_coords leading-blank-line stripping issue"). Left in place, it would have silently drifted from the now-fixed core splitter. Replaced it with `SdfRecordReader`, preserving the existing "silently skip malformed records" behavior. This also incidentally fixes a second, narrower bug in that loop: it matched `\$\$\$\$` as a substring anywhere (including inside a data-field value), where `SdfRecordReader`'s scan requires an entire line to equal `\$\$\$\$`.

Verified via `maturin develop --release` + pytest (new tests in `test_io.py` for both the blank-name fix and the malformed-skip contract; full `crates/chematic-py/tests/` suite re-run, no regressions attributable to this change).

Also disclosing, since it surfaced during that full pytest run and isn't otherwise being tracked: two **pre-existing, unrelated** test failures — `test_run_smirks_transfer_preserves_E`/`_Z` in `test_module_functions.py` — caused by v0.8.0's canonical-SMILES automorphism-orbit-aware branch pruning changing 2-butene's canonical spelling from `C/C=C/C` to the equivalent branched form `C(=C/C)\C`. Confirmed via direct `assign_cip()` comparison that the underlying E/Z stereo label is preserved correctly across the rewrite (not a correctness regression, just an assumed-fixed string the test hardcoded). Separately: `bash scripts/check.sh` and every CI workflow never run `pytest` at all, so this has been silently failing undetected since whatever merge introduced it. Filing both as a separate issue rather than fixing here — different root cause from this PR's scope.

Issue #227, opened 2026-08-01
chematic-ff: MMFF94 strict coverage gap — 216/265 molecules in Wave 1 benchmark unsupported, including plain benzene


kent-tokyo:
## Summary

Discovered during the Wave 1 pipeline v2 vs RDKit ETKDGv3 benchmark (PR #226), not fixed there per that PR's measurement-only scope. `ForceFieldPolicy::Mmff94BondAngleStrict` fails on **216/265 (81.5%)** of the benchmark's corpus, including molecules as simple as plain benzene, due to missing MMFF94 angle/torsion parameters.

This is a distinct root cause from issues #185/#188 (UFF minimizer convergence/blowup on specific ring counts) — this is a **parameter coverage gap**, not a minimizer convergence issue, so filed separately rather than folded in.

## Scope of the gap

- 216/265 molecules fail with `PipelineV2FailureCause::ForceField(MissingParameters(...))`.
- By corpus tier: 28/65 in Tier A (curated stress corpus), 188/200 in Tier B (ChEMBL-derived drug-like corpus).
- **100% of the 216 failures have at least one missing Angle parameter.** Missing-term counts, summed across all 216 failing molecules: Torsion 7084, Angle 5616, Bond 773, Oop 67.
- Affects basic aromatic chemistry, not just exotic elements: benzene, naphthalene, pyridine, thiophene, indole, purine, quinoline, anthracene, pyrene, and biphenyl are all in the failing set.

## Exact reproduction

```rust
use chematic_3d::pipeline_v2::{self as pv2, PipelineV2Config, StereoPolicy, RingTorsionApplicationPolicy};
use chematic_3d::distance_geometry_v2::EmbedParameters;
use chematic_3d::etkdg_knowledge::TorsionOptimizationConfig;
use chematic_3d::minimize::ForceFieldPolicy;

let mol = chematic_smiles::parse("c1ccccc1").unwrap(); // benzene
let config = PipelineV2Config {
    embed: EmbedParameters { random_seed: 20260801, max_attempts: 8, ..EmbedParameters::default() },
    torsion_optimization: TorsionOptimizationConfig::default(),
    include_legacy_torsion_heuristic: false,
    stereo_policy: StereoPolicy::Ignore,
    fail_on_unevaluable_stereo: false,
    force_field_policy: ForceFieldPolicy::Mmff94BondAngleStrict,
    force_field_max_iterations: 200,
    gate_mmff94_torsion_oop: false,
[...]

PR #248, opened 2026-08-04
bench(3d): re-run Wave 1 265-corpus benchmark on v0.11.0, add RepairAndVerify arms

kent-tokyo:
## Summary

Priority 1 of the post-#241 3D roadmap. Re-runs the full 265-molecule
`pipeline_v2` vs RDKit ETKDGv3 benchmark on v0.11.0 (`93aaa9e`), which
had not been re-run since `1bc1b63` (pre-v0.10.0) — so it never
reflected the MMFF94 typing fixes shipped in #236/#238/#239/#241.

- Adds 2 new **independent** arms (existing 5 `Ignore`-policy arms are
  unchanged): `chematic_pipeline_v2_mmff94_strict_repair` and
  `..._with_uff_fallback_repair`, both `StereoPolicy::RepairAndVerify`.
- Common scorer (`pipeline_v2_vs_rdkit_common_scorer.rs`) is
  byte-identical to the original `1bc1b63` version (`git diff` empty)
  — new vs. old numbers are apples-to-apples on the scoring side.
- Manifests unchanged (`corpus_sha256` verified identical before/after
  regeneration).
- Spot-checked `chematic_pipeline_v2_uff_only` (unchanged `Ignore`
  policy) against the pre-v0.11.0 baseline: 264/265 rows byte-identical;
  the 1 divergence (`macrolactam_12`) is explained by PR #245's
  macrocycle amide bound-pinning fix that shipped in v0.11.0, not by
  this refactor.
- New stage-funnel table follows `pipeline_v2`'s **real** execution
  order (stereo verify/repair happens *before* force-field
  minimization, not after) — caught and fixed a `>=`/`>` boundary bug
  before merging that had `ff_attempted`/`ff_succeeded` swapped.
- New RepairAndVerify paired-arm effectiveness table (before/after
  stereo mismatch, repair attempted/succeeded, geometry degradation,
  timing delta). Required extending `PipelineV2Failure`'s partial
  stereo evidence to failure rows in the dump script (previously
  emitted on success rows only) — molecules that reached repair but
  failed at a later stage were otherwise silently undercounted.
- Fixed several stale hardcoded numbers baked into the report's prose
  (a "216-molecule" gap figure, old arm counts, a one-off "615.3s"
  outlier anecdote from the original run) to compute dynamically from
  the aggregate instead.
- Process-level (whole-binary) perf intentionally **not** re-run this
  round: the arm matrix grew from 6 to 8, so the stored `1bc1b63`-era
  timing file would no longer measure the same binary. Moved aside
  rather than presented as if comparable; the in-process per-arm
  timing table is unaffected and remains the primary comparable metric
  this round.

No production chemistry code changed — only the benchmark example,
report generator, and result/doc data.

## Key numbers (v0.11.0, full pipeline scope)

- `mmff94_strict`: **38/265 → 149/265** full-pipeline success. This IS a
  direct, apples-to-apples comparison — verified against `93aaa9e`'s own
  report (`docs/pipeline_v2_vs_rdkit_etkdgv3_benchmark.md` at that commit):
  the `chematic_pipeline_v2_mmff94_strict` row there is the same arm name,
  same full `pipeline_v2`/`embed_pipeline_v2` harness, same 265-molecule
  corpus, same common scorer — just measured before the #236/#238/#239/#241
  MMFF94 typing fixes landed. This is the first full-pipeline re-measurement
  to reflect those fixes, and the 38→149 delta is directly attributable to
  them.
  What is genuinely NOT comparable to either 38/265 or 149/265: the
  `minimize_with_policy`-direct-call numbers cited in commit messages along
  the way (44/265, 130/265, from `mmff94_strict_gate_remeasure_227.rs`) —
  that script bypasses `pipeline_v2` entirely (bare `dg::generate_coords` +
  `minimize_with_policy`, no embed pipeline, no stereo, no stage tracking),
  a structurally different, narrower measurement than either full-pipeline
  number.
  Full-pipeline `mmff94_strict` also has 116/265 unsupported under issue
  #227, ~6,900 stretch-bend terms still ungated by the strict check at the
  time of this PR (re-measured and corrected to 2,107 in the Priority 2
  follow-up, PR #249).
- `mmff94_strict_repair` / `mmff94_with_uff_fallback_repair`: 131/265
  and 229/265 success respectively — lower than the matching `Ignore`
  arms (149/265, 252/265) because `RepairAndVerify` gates success on
  repair outcome — but both reach **100% stereo satisfaction among
  their successes**, matching RDKit on that subset.
- Legacy `etkdg`: 265/265 raw success, 251/265 (94.7%) independently
  sound under the stricter common scorer (14 coincident-atom cases).

Full breakdown: `docs/pipeline_v2_vs_rdkit_etkdgv3_benchmark.md`.

## Test plan

- [x] `bash scripts/check.sh` green (build, clippy, fmt, full test
      suite, cargo-deny, version consistency)
- [x] `pipeline_v2_vs_rdkit_common_scorer.rs` verified byte-identical
      to `1bc1b63` via `git diff`
- [x] Manifest `corpus_sha256` verified identical before/after
      regeneration
- [x] `chematic_pipeline_v2_uff_only` spot-checked against pre-v0.11.0
      baseline (264/265 byte-identical, 1 explained divergence)
- [x] Stage-funnel monotonicity self-checked (all 8 arms, all 7
      columns non-increasing)
- [x] Repair-effectiveness self-checked (`repair_succeeded <=
      repair_attempted`, `n_compared + n_excluded == 265` for both
      repair arms)
- [x] Report generator's own integrity gates (row-count,
      unclassified-rows, atom-mapping, missing/mismatched coords,
      non-finite coords, common-scorer coverage, denominator
      self-consistency) all pass at generation time

**Not merging without explicit go-ahead**, per this repo's standing
release/PR authorization pattern.


kent-tokyo:
Update: pushed a follow-up commit clarifying 3 things a reviewer would trip on:

1. **Reviewer note on the deleted file**: `validation/results/pipeline_v2_vs_rdkit_process_level_perf.json` is deleted in this PR (not just modified). It's the `1bc1b63`-era whole-binary timing (6 chematic arms); since this PR grows the arm matrix to 8, that file would no longer measure the same binary if kept, so it's removed rather than presented as comparable. The raw data is still recoverable from git history at `1bc1b63` — please confirm this deletion is fine, or ask if you'd rather it stay archived under a different filename instead.
2. Added an explicit note (verified per-molecule, not assumed) explaining why the two RepairAndVerify arms report identical before-mismatch/attempted/succeeded counts — structural, not a bug (repair runs pre-FF).
3. Reframed the force-field convergence conclusion row: 17.1% converged within 200 iterations but 252/265 runs still pass final validation — flagged as an input to Priority 3 (Stage 1C) rather than a flat RDKit-favor line.

kent-tokyo:
Self-review pass (no human reviewer assigned yet): found and fixed one real self-contradiction in the new Stage funnel section, in `d4affea`.

**Bug**: the Stage funnel table's note claimed `chematic_legacy_etkdg`'s intermediate columns are "0 by construction" (legacy has no `PipelineStage` tracking), but the table directly above it showed `265` for every column on that row. `stage_funnel()` gave every arm's success rows a `cutoff_idx` of `n_stages`, so legacy's 100%-raw-success rows trivially "passed" every stage check even though that code path never runs `DistanceGeometry`/`StereoRepair`/`ForceFieldMinimization` at all.

**Fix**: `chematic_legacy_etkdg` is now special-cased to report `n/a` for the intermediate columns (not a fabricated 0, not the misleading 265). Also tightened the funnel's blanket "reached" prose sentence -- it only described the `*_succeeded`/passed semantics (`cutoff_idx > idx(S)`), not the `*_attempted`/`*_reached` semantics (`cutoff_idx >= idx(S)`) used by 2 of the 7 columns -- worth being explicit about since this session already hit and fixed the exact same reached/passed boundary bug once (see the PR description).

Scope check: only `scripts/gen_pipeline_v2_vs_rdkit_report.py`, the regenerated `docs/...md`, and `validation/results/pipeline_v2_vs_rdkit_aggregate.json`'s `stage_funnel.chematic_legacy_etkdg` sub-object changed -- verified via `git diff` that every other number/hash in the aggregate is byte-identical. `bash scripts/check.sh` green.

Still not merging without explicit go-ahead.

PR #249, opened 2026-08-05
feat(ff): gate MMFF94 stretch-bend coverage, re-classify missing terms (Priority 2 / Stage 1B, issue #227)

kent-tokyo:
## Summary

Priority 2 (Stage 1B) of the post-#241 3D roadmap. Rebased onto `main`
now that #248 is merged.

Stretch-bend cross terms (Halgren MMFF.V eq. 4) were never counted toward
`Mmff94BondAngleStrict`/`Mmff94WithUffFallback`'s coverage gate at all —
`Mmff94CoverageReport` had no field for them, and `stretch_bend_energy`
(chematic-ff's `mmff94_minimizer`) silently contributes zero energy for
an uncovered term instead of erroring.

**This PR went through one review round already** (see the review
comment below) — 4 real blockers found and fixed, verified with a fresh
full-corpus re-run each time. The description below reflects the
corrected, current state.

### What changed (production code)

- `Mmff94TermKind::StretchBend` + `Mmff94CoverageReport::stretch_bend_{total,missing}`
  — `compute_mmff94_coverage` now measures it, mirroring the angle-bend
  loop's exact `(i,j,k)` triples.
- `include_stretch_bend_in_gate` / `gate_mmff94_stretch_bend` — an
  independent opt-in with the same shape as the pre-existing
  `gate_mmff94_torsion_oop` (`minimize_with_policy_gated`/
  `PipelineV2Config`). **Defaults to `false`, so no existing arm's
  behavior changes.**
- Threaded through `chematic-py` and `chematic-wasm` (WASM's JSON field
  is `#[serde(default)]` — old caller configs keep parsing; see
  `CHANGELOG.md` for the full, deliberately-asymmetric Rust/Python/WASM
  compatibility posture).
- **4 new independent** `pipeline_v2_vs_rdkit_dump.rs` benchmark arms — a
  real 3-stage comparison, never edits to existing arms:
  `legacy` (bond+angle) → `..._stretch_bend_gated` (+stretch-bend) →
  `..._complete_bonded_term_gated` (+torsion+OOP too), for both
  `mmff94_strict` and `mmff94_with_uff_fallback`.

## Measured (265-molecule corpus, same manifests/scorer as #248)

Fresh `mmff94_term_coverage_audit.rs` re-run: **2,107** missing
stretch-bend instances (not the stale ~6,900 in
`docs/mmff94_coverage_gap_227_audit.md`).

Per-term `routing_bug_candidate` (table row exists at a *different*
classification code) vs. `table_gap` (absent everywhere):

| Term kind | total | routing_bug_candidate | table_gap |
|---|---|---|---|
| Bond | 84 | 79 (94.0%) | 5 (6.0%) |
| Angle | 374 | 277 (74.1%) | 97 (25.9%) |
| Torsion | 1121 | 1107 (98.8%) | 14 (1.2%) |
| Oop | 0 | 0 | 0 |
| StretchBend | 2107 | 427 (20.3%) | **1680 (79.7%)** |

**New this round**: StretchBend's `table_gap` is further sub-classified
against RDKit's actual stretch-bend algorithm, verified from the pinned
RDKit commit's real source (`Code/GraphMol/ForceFieldHelpers/MMFF/AtomTyper.cpp`,
`Code/ForceField/MMFF/Params.{h,cpp}`) — confirmed the real fallback
chain is: specific/generic MMFF-type lookup, then `MMFFDfsbCollection`'s
periodic-table-row default (`defaultMMFFDfsb`, 29 rows, ported verbatim
as a diagnostic-only classifier, never wired into production) — and that
**no equivalence-class step exists anywhere in RDKit's real stretch-bend
path** (that mechanism is angle/torsion/OOP-only in RDKit itself).
Result: **100% (1680/1680) of the StretchBend `table_gap` would be
resolved** by porting this small table into production chematic-ff —
verified with negative controls (correctly returns `false` for
out-of-table combinations and the table's one all-zero row, not a
blanket `true`). This is a concrete, narrow, low-risk follow-up
candidate, not resolved in *this* PR.

**3-stage paired-arm comparison:**

| Stage transition | mmff94_strict | mmff94_with_uff_fallback |
|---|---|---|
| legacy → stretch_bend_gated | 148 → **37** (111 newly fail) | 252 → 250 (2 newly fail) |
| stretch_bend_gated → complete_bonded_term_gated | 37 → 37 (**0** newly fail) | 250 → 250 (**0** newly fail) |

The stretch-bend-gated and complete-bonded-term-gated numbers are
numerically identical in this corpus (verified empirically, not
assumed) — every molecule that survives the stretch-bend gate also has
complete torsion/OOP coverage here. **Not adopted as the new default** —
existing arms untouched.

`mmff94_with_uff_fallback`'s legacy success count moved 251→252 between
an earlier draft and this run: `chembl_tier_b_0166`'s MMFF94 attempt
sits right at the 20s wall-clock timeout boundary (18710ms vs.
21378ms) — `embed_seed` governs geometry determinism, not real-time
scheduling, so this specific molecule can legitimately flip across runs
independent of any code change. 252 matches the count already on `main`
(via #248). Documented in the report so a future 251/253 isn't mistaken
for a regression.

## Scoped out of this PR

The originally-proposed 4-way exact-missing/equivalence-fallback-not-
implemented/empirical-rule-not-implemented/genuinely-unsupported
classification is **not** fully implemented for Bond/Angle/Torsion/Oop —
chematic-ff implements neither MMFF94 equivalence-class substitution nor
empirical-rule estimation at all for those term kinds
(`Mmff94NumericTypeInfo.equivalence_levels` carries real data but has
zero readers anywhere in the codebase, verified). For StretchBend
specifically, this is now resolved differently than originally planned:
RDKit's own algorithm doesn't use equivalence-class substitution for
stretch-bend at all, so the Dfsb classification above **is** RDKit's
complete residual-fallback story for that term kind, not a partial one.

## Test plan

- [x] `bash scripts/check.sh` green (build, clippy, fmt, full Rust +
      667-test Python suite, cargo-deny, version consistency)
- [x] `mmff94_with_uff_fallback_falls_back_and_reports_why_on_chfclbr`
      updated (`missing_parameter_classes` 3 → 6, verified not
      double-counted)
- [x] Stage-transition invariant enforced at report-generation time:
      `Mmff94BondAngleStrict` (pure gate) asserted monotonic across both
      transitions; `Mmff94WithUffFallback` exceptions independently
      verified against 7 specific fields (status/failure_cause/
      failure_stage on the earlier row; status/force_field_actual/
      fallback/fallback_reason/coverage-evidence on the later row) before
      being accepted, not asserted away with a weaker check
- [x] Dfsb port verified with negative controls (out-of-table periodic-row
      combinations and the table's one all-zero row both correctly
      return `false`)
- [x] WASM backward-compat regression test added
      (`pre_priority2_config_json_without_gate_stretch_bend_still_parses`)
- [x] RDKit rows/manifests/`common_scorer.rs` reused unchanged
      (chematic-only change)

**Not merging without explicit go-ahead**, per this repo's standing
release/PR authorization pattern.


kent-tokyo:
Addressed all 4 blockers from the review, commit \`5373b11\`. Summary (full detail in the updated PR description / \`docs/pipeline_v2_vs_rdkit_etkdgv3_benchmark.md\`):

1. **37/265 mislabeled "true complete-term"** — added 2 more independent arms (\`gate_torsion_oop=true\` AND \`gate_stretch_bend=true\`), forming a real 3-stage legacy → stretch-bend-gated → complete-bonded-term-gated comparison. Measured (not assumed): the last transition is 0-newly-failing for both policies in this corpus — reported as an empirical, corpus-specific finding.

2. **1,680 "table_gap" called genuine without checking RDKit's real fallback** — fetched the pinned RDKit commit's actual source and confirmed the real algorithm: specific/generic lookup → \`MMFFDfsbCollection\`'s periodic-table-row default (29-row table) → nothing else (no equivalence-class step exists for stretch-bend in RDKit at all — that's angle/torsion/OOP-only). Ported the table verbatim as a diagnostic-only classifier, verified with negative controls, result: 100% (1680/1680) resolvable by porting it into production. Reframes "unclear next step" into a concrete, small, low-risk follow-up.

3. **Timeout-rescue check weaker than its comment claimed** — strengthened from a single \`status==timeout\` check to 7 verified fields across both rows (failure_cause/failure_stage on the earlier row; force_field_actual/fallback/fallback_reason/coverage-evidence on the later row).

4. **"default false" wasn't API-compatible** — WASM's JSON field now has \`#[serde(default)]\` (regression-tested); Python's \`.safe()\` already had a default, \`.new()\` deliberately left without one (matches its own "no hidden default" docstring and \`gate_mmff94_torsion_oop\`'s existing precedent). CHANGELOG migration note added, minor version bump recommended.

Also caught and documented while re-measuring: a run-to-run wall-clock timeout boundary flip (\`chembl_tier_b_0166\`, explained not hidden), and a stale "~6,900" figure earlier in the same generated report that this PR's own fresh measurement (2,107) contradicted.

CI is running on the fresh push. Still not merging without explicit go-ahead.

PR #250, opened 2026-08-05
feat(ff): port RDKit's Dfsb periodic-row stretch-bend fallback into production (Priority 2B, issue #227)

kent-tokyo:
## Summary

Priority 2B of the post-#241 3D roadmap (issue #227). Follows the order
the user specified after reviewing #249: **#248/#249 merged → Priority
2B (this PR) → re-measure → Priority 3**, since diagnosing
MinimizationFailed (Priority 3) before this PR would diagnose a
population that immediately shifts once more molecules newly reach the
minimizer.

Priority 2 (#249)'s own diagnostic audit found the single largest
missing-term bucket (StretchBend `table_gap`, 1,680 instances) was
**100% resolvable** by porting a small, pinned-RDKit-commit-verified
29-row periodic-table-row fallback table (`MMFFDfsbCollection`'s real
RDKit equivalent) into production. This PR ships that port — scoped
narrowly per the reviewer's explicit instruction: **does not touch
bond/angle/torsion routing-bug candidates**, single root cause only.

## What changed

- `chematic_ff::mmff94_stbn` now tries the existing specific/generic
  MMFF-type table first (unchanged, always wins if it has a row), then
  falls back to the ported Dfsb table on failure — **unconditional
  production behavior, not behind any opt-in flag** (a production
  accuracy fix, not a diagnostic feature). Applies to every MMFF94
  policy's energy/gradient calculation (`stretch_bend_energy`) and to
  `Mmff94CoverageReport`'s stretch-bend coverage measurement
  (`compute_mmff94_coverage`) the same way.
- The prior type-only lookup remains available as the new
  `mmff94_stbn_type_only` (used by diagnostic tooling's "present at a
  different classification code" check, which is independent of any
  element-based fallback — verified no equivalence-class step exists in
  RDKit's real stretch-bend path either, per #249's own research).
- `mmff94_stbn`'s public signature gained 3 required
  `atomic_num_{i,j,k}: u8` params (the periodic-row lookup is
  element-keyed, not MMFF-type-keyed) — breaking for external callers,
  documented in `CHANGELOG.md`.
- **`gate_mmff94_stretch_bend`/`gate_mmff94_torsion_oop` (Priority 2's
  strict-refusal gates) are UNCHANGED**, remain independent opt-ins,
  still `false` by default — this PR only changes what counts as
  "covered" underneath those gates, not whether the gates are on.
- `mmff94_term_coverage_audit.rs`'s diagnostic-only Dfsb duplicate
  (added in Priority 2 to answer "would this resolve if ported") is
  removed. The audit's StretchBend check is now a **two-axis design**
  instead: a TYPE-ONLY diagnostic axis (`mmff94_stbn_type_only`,
  independent of Dfsb — still reports every classification-routing
  candidate) and a FINAL PRODUCTION-RESOLUTION axis (`dfsb_resolved`,
  the new production `mmff94_stbn`) tracked separately per row, so
  Dfsb's own success can never silently overwrite the diagnostic signal
  that motivated building it (see Measured, below).

## Measured (265-molecule corpus, same manifests/scorer as #248/#249)

- Missing stretch-bend instances (type-only diagnostic axis): 2,107,
  unchanged by design. Final production-unresolved: **2,107 → 0**
  (confirms Priority 2's diagnostic finding end-to-end, not left as an
  untested "would" claim).
  - Of those 2,107: **1,680 are genuine table gaps**, now correctly
    closed by Dfsb (matches RDKit's own behavior for these triples).
  - The remaining **427 are routing-bug candidates** — a real,
    correctly-typed parameter exists at a *different* MMFF
    classification code than the one actually computed. Dfsb's generic
    periodic-row fallback also happens to resolve these, but that's
    coverage, not correctness: chematic returns RDKit's generic
    default instead of the specific, correctly-routed parameter for
    these 427. **Not fixed in this PR** — masked, and called out
    explicitly rather than silently absorbed into the "0 missing"
    headline number. Left for a separate follow-up issue.
- `mmff94_strict_stretch_bend_gated`: 37 → **149/265**, converges
  EXACTLY to legacy `mmff94_strict`'s success set (verified: identical
  molecule sets via a per-molecule set-equality check, not just
  identical counts) — the stretch-bend gate essentially never fires
  anymore.
- `mmff94_with_uff_fallback_stretch_bend_gated`: 250 → 252/265, same
  convergence to its legacy counterpart.
- `mmff94_strict_complete_bonded_term_gated` (still gates torsion/OOP,
  untouched by this fix): 37 → **86/265** — the residual gap is a
  separate, known issue (torsion coverage, 98.8%
  routing-bug-candidate per Priority 2's own classification — likely a
  classification-derivation bug, not a table gap), explicitly not
  addressed in this PR.
- **Regression check**: full per-molecule diff against the
  pre-Priority-2B baseline (saved before re-running) shows **0
  molecules lost** under any gated arm — only gains. The legacy
  `mmff94_strict`/`mmff94_with_uff_fallback` arms are **not** claimed
  output-invariant: gate *eligibility* is unchanged (neither arm ever
  gated stretch-bend, before or after this PR), but Dfsb changes every
  MMFF94 policy's energy/gradient unconditionally, which can shift
  minimizer convergence and therefore success/failure. The one observed
  change (`mmff94_strict` 148 → 149) was traced via per-molecule diff to
  `chembl_tier_b_0166` (`elapsed_ms` 20530 → 16221, status
  timeout → success) — a net gain, verified with real evidence rather
  than assumed, and reported as such rather than asserted to be pure
  wall-clock jitter without checking.

| Arm | before | after |
|---|---|---|
| `mmff94_strict_stretch_bend_gated` | 37/265 | 149/265 |
| `mmff94_with_uff_fallback_stretch_bend_gated` | 250/265 | 252/265 |
| `mmff94_strict_complete_bonded_term_gated` | 37/265 | 86/265 |
| `mmff94_with_uff_fallback_complete_bonded_term_gated` | 250/265 | 250/265 |

## Test plan

- [x] `bash scripts/check.sh` green (build, clippy, fmt, full Rust +
      667-test Python suite, cargo-deny, version consistency)
- [x] New unit tests (`crates/chematic-ff/src/mmff94_energy/mod.rs`):
      type-table hit always wins over Dfsb; Dfsb resolves a real
      F-C-Cl-style triple with a value traceable to a specific
      pinned-source row; swap/canonicalization correctness; the
      table's one all-zero row stays unresolved (matches RDKit's own
      `isDoubleZero` exclusion); a structurally-impossible row_j=0
      case stays unresolved
- [x] `oop_stbn.rs`: table has exactly 29 rows (regression guard), all
      rows canonically `row_i <= row_k`
- [x] `mmff94_with_uff_fallback_falls_back_and_reports_why_on_chfclbr`
      updated: CHFClBr's 3 halogen-C-halogen stretch-bend triples now
      resolve via Dfsb (F/Cl/Br/C all within covered periodic rows) —
      `missing_parameter_classes` 6 → 3, `stretch_bend_missing` 3 → 0
- [x] Full 265-molecule × 12-arm corpus re-run from scratch (not reused)
- [x] Per-molecule regression diff against the pre-this-PR baseline —
      0 unexpected status/soundness changes

Recommends **v0.12.0** (minor), consistent with #249's own breaking
struct-field addition already requiring the same bump — confirmed by
the reviewer, not self-assessed.

## Review round

A prior version of this PR (`870b94d`) was held back from merge after
review found 2 real blockers in the audit tool / report, both fixed in
`89fbc8f`:

1. The audit tool's stretch-bend check originally only fired on the
   FINAL (Dfsb-inclusive) lookup, silently discarding the 427-instance
   routing-bug-candidate signal Priority 2 had found. Fixed via the
   two-axis design described above.
2. The report/CHANGELOG claimed legacy arms' success sets were
   "structurally unchanged" — contradicted by the data's own 148 → 149
   delta. Fixed by stating the real invariant (gate eligibility, not
   output) and citing verified per-molecule evidence for the delta.

All 18 CI checks passed on the first attempt against `89fbc8f`.

**Merge approved** by explicit reviewer go-ahead after confirming both
fixes above.


kent-tokyo:
CI note: the first run's \`Rust Criterion regression gate\` failed on \`pka_acid_base_10mol\` (median ratio 1.0557 vs. the 1.04 practical-effect threshold, 10 paired trials, flagged \`tiny_sample: true\`). This benchmark measures pKa/SMARTS descriptor calculation in \`chematic-chem\` — completely unrelated to this PR's \`chematic-ff\`/\`chematic-3d\` MMFF94 stretch-bend changes, and consistent with this gate's known pseudo-replication measurement issue (issue #70: ~100 Criterion samples/side aren't independent trials, since they share one process's runner state — a small environment difference can amplify into a spurious "regression" on an unrelated benchmark). Re-ran the job for an independent sample (\`gh run rerun --failed\`, not a threshold/gate change) — it now passes. All 18 checks green.

kent-tokyo:
Addressed both blockers from the review, commit \`89fbc8f\`.

**Blocker 1 (427 routing candidates disappeared, not fixed)**: \`mmff94_term_coverage_audit.rs\` now emits a row whenever the TYPE-ONLY lookup misses, regardless of whether Dfsb then rescues it, with a new \`dfsb_resolved\` field alongside \`present_at_different_classification\` on the same row. Restored the full split (verified against the raw JSONL, not assumed):
- \`dfsb_resolved_true_type_table_gap\` = 1,680 (genuine gaps, Dfsb closing these matches RDKit's real behavior)
- \`dfsb_resolved_routing_candidate\` = 427 (masked, not fixed — chematic now uses RDKit's generic default instead of the correctly-routed specific parameter)
- \`final_unresolved\` = 0

Report/summary JSON updated to keep these two axes (type-only diagnostic vs. final production resolution) permanently separate, with an explicit "not fixed in this PR, different root cause" note. Did not touch the 427 or any bond/angle/torsion routing candidates — single root cause preserved.

**Blocker 2 ("legacy success set structurally unchanged" claim)**: rewrote the report/CHANGELOG to state the real invariant (bond+angle *gate eligibility* unchanged) rather than an output-invariance claim the data itself contradicted. Traced the 148→149 \`mmff94_strict\` delta to a specific molecule via the saved pre-PR per-molecule baseline: \`chembl_tier_b_0166\`, \`elapsed_ms\` 20530→16221, \`status\` timeout→success. Reported as verified-but-not-fully-explained (a ~4.3s drop is a substantial, consistent-direction change, not obviously pure machine-load noise) rather than asserted as "known jitter" without checking — noted that the same molecule ID also recurred as the timeout-boundary case in Priority 2's own measurement. Also fixed CHANGELOG's "(unaffected-by-gating)" parenthetical, which read as an output-invariance claim.

\`bash scripts/check.sh\` green. Still not merging without explicit go-ahead.

Issue #139, opened 2026-07-22
VF2: prune automorphic branches for symmetric SMARTS targets


kent-tokyo:
## Background

PR #138 shipped a genuine three-way `MatchOutcome::{Found, NotFound, BudgetExhausted}` for PAINS/Brenk substructure matching, replacing an unbounded VF2 search that could hang indefinitely on highly symmetric targets. `BudgetExhausted` is conservatively folded to "alert present" (flagged) by the existing non-`_checked` API, matching the project's fail-safe design.

That fix is correct and intentionally does not solve the underlying algorithmic problem: VF2's backtracking blows up combinatorially on local automorphisms (e.g. tert-butyl / gem-dimethyl symmetry), so some ordinary, non-exotic molecules resolve to a false-positive PAINS flag purely because proving the true negative costs more than the shipped budget affords.

## Known concrete case

- Fixture: a di-tert-butylphenol scaffold, PAINS rule `tert_butyl_B(1)`.
- True answer: `NotFound`.
- At the shipped budget (1,000,000 visits): resolves to `BudgetExhausted`, folded to "flagged" by the current non-`_checked` API (accepted, documented tradeoff — see PR #138).
- At 20,000,000 visits: correctly resolves to `NotFound`, but costs ~20 seconds for this one molecule/pattern pair.

## Goal

Not "raise the budget" — that only pushes the same tradeoff to a different (unknown) breaking point, and costs real wall-clock time even where it does work. The real fix is symmetry-aware candidate ordering or automorphism pruning in the VF2 implementation itself (`chematic-smarts::match_vf2`), so the search doesn't re-explore symmetric branches that are already known to be equivalent.

## Acceptance criteria

- The `tert_butyl_B(1)` / di-tert-butylphenol fixture resolves to a confirmed `NotFound` at the shipped budget (no budget increase required).
- No slowdown on the normal (non-symmetric) corpus.
- The three-way `MatchOutcome` API and its conservative fold policy are preserved — this is an internal search-efficiency improvement, not an API/behavior change.

## Non-goal

Raising `ALERT_MAX_VISIT_BUDGET`. That is explicitly not the fix being tracked here.

kent-tokyo:
Resolved in the current checkout.

- VF2 now selects the next query atom by mapped-neighbor count, then query degree, with a deterministic index tie-break. This is an exhaustive traversal-order change; mappings and the typed MatchOutcome budget contract are unchanged.
[...]

PR #257, opened 2026-08-08
feat(3d): symmetry-aware Kabsch RMSD (rmsd_symmetric), oracle-validated

kent-tokyo:
## Summary

Priority 4 groundwork: adds `chematic_3d::conformer::rmsd_symmetric`, a symmetry-aware Kabsch RMSD ported from RDKit's `GetBestRMS` (`Code/GraphMol/MolAlign/AlignMolecules.cpp`). Enumerates the molecule's own graph automorphisms (VF2 self-match, `uniquify: false`) and returns the minimum Kabsch-aligned RMSD over every automorphism-consistent atom relabelling, instead of a fixed atom-index correspondence.

This matters for the upcoming reference-conformer benchmark (Priority 4 in the roadmap): a plain fixed-index RMSD is wrong on any molecule with permutation-equivalent atoms (terminal `-CF3`'s three fluorines, neopentane's four methyls, etc.) — it would report a large spurious distance for conformers that are physically identical up to a relabelling of interchangeable atoms.

## Oracle validation

New `crates/chematic-3d/examples/rmsd_symmetric_oracle_dump.rs` + `scripts/rmsd_symmetric_oracle_check.py` independently cross-check chematic's result against RDKit's `rdMolAlign.GetBestRMS` on 6 cases (propane control, `-CF3`-ethane, neopentane, benzene, acetate, ibuprofen control).

- **5/6 agree with RDKit to within 2e-6 Å.**
- Benzene's nonzero residual (1.143095 Å, identically in both engines) is *correct*, not noise — verified by independent derivation: swapping two *para ring* carbons' positions is not a graph automorphism of a plain hexagon (unlike swapping two leaf atoms on the same parent, e.g. `-CF3`'s fluorines or neopentane's methyls), so no relabelling can recover 0.
- **Known, documented gap:** acetate disagrees (chematic 0.356 Å vs. RDKit 0.0 Å). RDKit's `GetBestRMS` additionally runs `symmetrizeConjugatedTerminalGroups`, treating a carboxylate's two oxygens as interchangeable despite differing formal bond order. `rmsd_symmetric` does not port that preprocessing step — disclosed in its own doc comment, not silently swept under a passing test. Tracked as a future follow-up, not a blocker for Priority 4's benchmark groundwork.

Full write-up: `validation/README.md`, new "Symmetric RMSD oracle check" section.

## Scope

- No production chemistry behavior changed outside the new `rmsd_symmetric` function and `ConformerEnsemble::conformer_rmsd_symmetric` convenience wrapper — both purely additive (new `pub fn`s, new `pub use` in `lib.rs`).
- Does not touch `dg.rs`, MMFF94, or any code from the recently-merged #253/#254.

## Test plan

- [x] `bash scripts/check.sh` green (fmt, clippy `-D warnings`, full workspace test suite, cargo-deny, publish graph, version consistency)
- [x] New unit tests in `conformer.rs`: self-RMSD is 0, symmetric RMSD never exceeds fixed-index RMSD, symmetric RMSD recovers ~0 on an automorphism-only CF3 fluorine swap where fixed-index RMSD is nonzero, `ConformerEnsemble` wrapper matches the free function, out-of-range index returns `None`
- [x] Independent RDKit oracle check (`scripts/rmsd_symmetric_oracle_check.py`) — 5/6 exact agreement, 1/6 documented known gap

Not merging — opening Ready per this session's established go-ahead pattern.

### DIFF
--- a/crates/chematic-3d/examples/rmsd_symmetric_oracle_dump.rs
+++ b/crates/chematic-3d/examples/rmsd_symmetric_oracle_dump.rs
@@ -1,0 +1,97 @@
+//! Dumps conformer pairs + chematic's `rmsd_symmetric` result for a handful
+//! of molecules chosen to exercise graph-automorphism symmetry (terminal
+//! -CF3, neopentane's 4 methyls, benzene's ring, a carboxylate to
+//! demonstrate the documented `symmetrizeConjugatedTerminalGroups` gap) plus
+//! one plain drug-like molecule with no interesting symmetry as a control.
+//!
+//! One conformer is `dg::generate_coords`'s output; the second is that same
+//! conformer under a fixed rigid rotation+translation, with two
+//! automorphism-equivalent atoms' positions additionally swapped where the
+//! molecule has any (so the comparison genuinely exercises alignment AND
+//! symmetry matching together, not a trivial self-comparison).
+//!
+//! Run:
+//! ```text
+//! cargo run --release -p chematic-3d --example rmsd_symmetric_oracle_dump \
+//!   > /tmp/rmsd_symmetric_oracle_dump.jsonl
+//! .venv/bin/python scripts/rmsd_symmetric_oracle_check.py \
+//!   /tmp/rmsd_symmetric_oracle_dump.jsonl
+//! ```
+
+use chematic_3d::conformer::rmsd_symmetric;
+use chematic_3d::coords::{Coords3D, Point3};
+use chematic_3d::dg::generate_coords;
+use chematic_core::AtomIdx;
+use serde_json::json;
+
+/// name, SMILES, (optional) pair of 0-based atom indices to swap in the
+/// second conformer (symmetry-equivalent atoms, chosen by inspection of the
+/// SMILES atom order).
+type Case = (&'static str, &'static str, Option<(usize, usize)>);
+const CASES: &[Case] = &[
+    ("propane", "CCC", None),
+    ("cf3_ethane", "FC(F)(F)C", Some((0, 2))), // two of the three F's
+    ("neopentane", "CC(C)(C)C", Some((0, 2))), // two of the four methyl carbons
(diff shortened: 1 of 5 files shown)


kent-tokyo:
Status update (2026-08-11): not merging as-is, but not abandoned either — this now feeds directly into the v0.14.0 plan.

**Why not merge as-is**: `rmsd_symmetric`'s current automorphism enumeration is unbounded — no `LimitExceeded`-shaped outcome, no bound on search cost for highly symmetric structures. Shipping it as a public API in this shape risks a pathological-input hang, and it would also grow a second, divergent symmetry-determination code path alongside #272's.

**Where it's going**: v0.14.0's A-tier item 3 ("exact bounded symmetry engine") is a single shared kernel — `exact_orbits()`, `{Exact, LimitExceeded}`-typed — meant to serve #257, #272, meso-compound detection, #139's VF2 automorphic-branch pruning, and the new production conformer-ensemble API's symmetry-aware RMSD dedup step, all at once, instead of each growing its own. `rmsd_symmetric` becomes the consumer that calls into that shared kernel rather than doing its own enumeration.

**Design input worth reading before the redesign** (added to ROADMAP.md 2026-08-10, from an unrelated ChemRxiv review): *StereoMolGraph* (ChemRxiv 2025-12-28, JCIM 2026-03-27) treats topological equivalence and stereochemical equivalence as genuinely separate notions, with both stereo-preserving and stereo-blind orbit computations exposed rather than one automorphism concept with stereo bolted on as a tie-break. `rmsd_symmetric` specifically should distinguish "topological symmetry" from "stereochemistry-preserving symmetry" as two different things a caller might want, not conflate them into one number — this PR's current form doesn't make that distinction.

Keeping this PR open; will pick it back up once the shared kernel work starts. No action needed from here in the meantime.

PR #259, opened 2026-08-09
feat(3d): add diagnostic-only torsion motif angle distribution analysis

kent-tokyo:
## Summary
- Adds a diagnostic-only Torsion Motif Angle Distribution (TMAD) analysis tool to `chematic-3d` (`src/torsion_motif.rs`): extracts rotatable-bond torsions from a molecule, builds a circular (wrap-aware) histogram of observed dihedral angles across a `ConformerEnsemble`, fits a von Mises mixture to that histogram, computes Jensen-Shannon divergence between two fitted profiles, tags a profile with a caller-supplied `TorsionEnvironment` label, and exports a deterministic JSON representation.
- Reuses `chematic-chem`'s existing RDKit-compatible rotatable-bond definition rather than inventing a second one: factored the shared filter predicate out of `rotatable_bond_count_from_set` into a new `is_rotatable_bond` helper, and added `rotatable_bond_atom_pairs(mol) -> Vec<(AtomIdx, AtomIdx)>` as a second caller of that same predicate. `rotatable_bond_count`'s behavior/callers are unchanged — same filter, still returns a `usize`, just no longer duplicates the exclusion logic that the new atom-pairs function also needs.
- Reuses the crate's existing `get_dihedral_deg` (mol_transforms.rs) for angle computation and `ConformerEnsemble` (conformer.rs) for multi-conformer iteration — no new geometry primitives.

**Diagnostic only**: nothing here touches ETKDG or any embedding/sampling code path. `TorsionEnvironment` is pure caller-supplied metadata — this crate has no model of what "crystal"/"vacuum"/"water"/"hexane" physically mean, it's just a label used to partition angle observations before comparison. This PR is a measurement tool for a future benchmark, not a generator change.

## Known scope limits (disclosed, not silently shipped)
`extract_torsion_motifs` reuses `chematic_chem::rotatable_bond_atom_pairs`'s existing "rotatable" definition rather than inventing a second, possibly-inconsistent one. Two consequences worth knowing about before using this for a benchmark, both documented in the module doc and pinned by tests (`amide_c_n_bond_is_not_a_torsion_motif`, `biphenyl_inter_ring_bond_is_a_motif_only_when_spelled_as_explicit_single`):
- **Amide C–N (omega) torsions are out of scope.** They're excluded from "rotatable" by the shared definition. A future PR wanting omega torsions would need a different extraction path (`chematic-3d`'s own `etkdg_knowledge::classify::candidate_central_bonds`/`classify_bond` keeps amide bonds and flags them via `amide_like`, but that module is currently private to this crate).
- **A biphenyl-like inter-ring bond is only found if spelled as an explicit single bond** in the input SMILES (`c1ccc(-c2ccccc2)cc1`). Without the explicit `-`, this crate's SMILES parser assigns `BondOrder::Aromatic` to that bond whenever both endpoints are lowercase, regardless of ring membership — so it's invisible to the "is a single bond" rotatable-bond filter. This is pre-existing parser/descriptor behavior, not something introduced or worked around by this PR.

## Attribution
Inspired by environment-conditioned torsion-distribution analysis in recent cheminformatics literature (circular-statistics histogram fitting + automated distribution comparison of torsion angles across different chemical environments, e.g. crystal packing vs. vacuum vs. solvent). The general ideas cited (circular histograms, mixture-of-von-Mises fitting, a distribution-distance metric) are standard circular-statistics tools, not verbatim from any paper — this is an independent from-scratch implementation:
- Von Mises density / EM mixture fit: method-of-moments M-step using the standard circular mean + Fisher's (1993, *Statistical Analysis of Circular Data*) closed-form kappa-inversion approximation.
- Modified Bessel function I0 (von Mises's normalizing constant): the classic Abramowitz & Stegun 9.8.1/9.8.2 polynomial approximation.
- Jensen-Shannon divergence: standard closed-form KL-divergence-based formula, implemented directly (no external crate).
- Test fixtures (synthetic von Mises samples) generated with an independently-implemented rejection sampler, not against the paper's own numbers (not available to this implementation).

No code, pseudocode, or verbatim text was ported from any paper.

## Test plan
- [x] `cargo test -p chematic-3d --lib torsion_motif` — 18 unit tests, all passing:
  - Circular histogram: observation-count preservation, boundary angles (±180°, ±360°, 540°) don't panic/go out of range, circular mean correctly resolves a cluster straddling the ±180° wrap boundary (verified it lands near 180°, not pulled toward 0° by a naive arithmetic mean), empty-histogram edge case, probabilities sum to 1.
  - Von Mises mixture fit, **both K=1 and K=2**: single-component EM fit recovers a known synthetic VonMises(mu=45°, kappa=4.0) sample's parameters (mu within 8°, kappa within 40% relative, 5000 samples, seeded deterministically). Two-component fit recovers a 60/40-weighted synthetic mixture of VonMises(-60°, kappa=8) and VonMises(60°, kappa=8) — sorted by mu before comparison since EM has no canonical component ordering, both means within 10° and both weights within 0.08 of ground truth. Both validated against synthetic data via an independently-implemented rejection sampler, not against any paper's own numbers (not available to this implementation). Empty-histogram fit returns no components.
  - Jensen-Shannon distance: `distance(a,a) == 0`, symmetric (`distance(a,b) == distance(b,a)`), bounded in `[0, ln(2)]`, and well-separated concentrated distributions approach the ln(2) bound.
  - Deterministic JSON export: identical inputs produce byte-identical JSON strings across two calls; exported JSON parses and round-trips its motif/environment/histogram/fit fields.
  - `extract_torsion_motifs`: butane finds its one rotatable-bond motif with the expected atom indices/signature; benzene (all ring bonds) finds none; amide C-N is confirmed absent; biphenyl's inter-ring bond is confirmed found only under the explicit-single-bond SMILES spelling (see "Known scope limits" above).
- [x] `bash scripts/check.sh` passes clean (fmt, clippy -D warnings, full workspace `--lib` and `--tests` suites, publish-graph, version-consistency).

Not merging — per repo convention this PR stays open/Ready pending explicit maintainer go-ahead.

### DIFF
--- a/crates/chematic-3d/Cargo.toml
+++ b/crates/chematic-3d/Cargo.toml
@@ -22,6 +22,7 @@ chematic-ff         = { path = "../chematic-ff",         version = "0.14.1" }
 chematic-chem       = { path = "../chematic-chem",       version = "0.14.1" }
 chematic-fp         = { path = "../chematic-fp",         version = "0.14.1" }
 chematic-smarts     = { path = "../chematic-smarts",     version = "0.14.1" }
+serde_json          = "1.0"
 
 # std::time::Instant panics on wasm32-unknown-unknown (issue #219); web-time
 # is a drop-in Instant replacement backed by Performance.now() there, and
@@ -34,4 +35,3 @@ web-time = "1"
 # comment -- a version-pinned dev-dependency must resolve against the
 # registry at `cargo publish` time, which a forward reference can't satisfy.
 chematic-smiles = { path = "../chematic-smiles" }
-serde_json       = "1.0"
--- a/crates/chematic-3d/src/lib.rs
+++ b/crates/chematic-3d/src/lib.rs
@@ -36,6 +36,7 @@ pub mod shape_descriptors;
 pub mod spectrophores;
 pub mod stereo3d;
 pub mod stereo_constraints;
+pub mod torsion_motif;
 pub mod usr;
 pub mod xyz;
 
@@ -89,6 +90,11 @@ pub use spectrophores::{
     SpectrophoresConfig, SpectrophoresNorm, spectrophores, tanimoto_spectrophores,
 };
 pub use stereo3d::{StereoAssignment3D, assign_stereo_from_3d};
+pub use torsion_motif::{
+    TorsionEnvironment, TorsionHistogram, TorsionMotif, TorsionProfileFit, VonMisesComponent,
+    extract_torsion_motifs, fit_von_mises_mixture, motif_angles_deg, torsion_profile_distance,
+    torsion_profile_to_json,
+};
 pub use usr::{shape_screen, usr_descriptors, usr_from_dg, usr_similarity};
 pub use xyz::{XyzError, parse_xyz, write_xyz};
 
--- a/crates/chematic-3d/src/torsion_motif.rs
+++ b/crates/chematic-3d/src/torsion_motif.rs
@@ -1,0 +1,892 @@
+//! Torsion Motif Angle Distribution (TMAD) — diagnostic-only.
+//!
+//! Extracts rotatable-bond torsions from a molecule, builds a circular
+//! (wrap-aware) histogram of their observed dihedral angles across a
+//! [`ConformerEnsemble`], fits a von Mises mixture to that histogram, and
+//! compares two fitted profiles with Jensen-Shannon divergence. Environment
+//! labels ([`TorsionEnvironment`]) are caller-supplied metadata only — this
+//! module has no notion of what "crystal" or "water" physically means, it
+//! just partitions/tags observations by whatever label the caller attaches
+//! to an ensemble.
+//!
+//! Inspired by environment-conditioned torsion-distribution analysis in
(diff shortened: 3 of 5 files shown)


kent-tokyo:
Status update (2026-08-11): not part of the v0.14.0 plan (which is scoped to stereo-constrained embedding, a production conformer-ensemble API, the shared symmetry engine, and MMFF94 completion) — this stays parked independent of that work, not superseded by it.

Outstanding items before this can merge, unchanged from the earlier review:
- The motif identity needs to be topology-only and atom-order-independent — it's currently sensitive to implicit-vs-explicit bond notation, so chemically identical inputs can land in different buckets depending on how they were written.
- Panics and silent-empty-string failures need to become typed errors.
- `torsion_profile_distance` should be renamed to `js_divergence` — it computes a Jensen-Shannon divergence, not a metric (JS divergence isn't a true distance metric; the sqrt would be, but that's not what's implemented), and the current name overclaims.

Keeping this PR open. Whoever picks it back up should treat the three items above as the acceptance bar, not just CI-green.

PR #261, opened 2026-08-09
spike(fp): reproduce superimposed count-fingerprint coding vs folding baselines

kent-tokyo:
## What this is

A **bounded reproducibility SPIKE** — research/exploration only, explicitly **not** a production feature request. It does not touch, wrap, or get called from `ecfp4()`, `mol.ecfp4()`, `rdkit_morgan_ecfp4_experimental()`, or any other public fingerprint API. The two existing ECFP4 engines are untouched (see confirmation below). **Do not merge.**

## What it investigates

"Superimposed coding": folding a *count* fingerprint (`&[(feature_hash, count)]`) down to a fixed-width *binary* vector such that binary Tanimoto on the folded bits approximates the true multiset (count-aware) Tanimoto better than plain presence folding or RDKit-style count-simulation folding alone.

New standalone module: `crates/chematic-fp/src/superimposed_coding.rs` (not re-exported at the crate root — reachable only via `chematic_fp::superimposed_coding::*`, same pattern as the existing `diagnostics` module). Provides, for comparison, on the same synthetic inputs:

- `fold_presence_counts` — plain presence/absence folding, counts discarded.
- `count_simulation_fold` — RDKit-style count simulation: up to `max_repeats` bits per feature, one per unit of count, at deterministic pseudo-random positions.
- `superimposed_code_counts` — this spike's design: count encoded as a unary/thermometer ladder of up to `repetitions` layers (layer `l` active iff `count > l`), each active layer contributing its own `codeword_weight`-bit sparse codeword OR'd into the shared vector.
- `exact_count_tanimoto` — exact multiset Tanimoto (`sum(min)/sum(max)`) on the raw unfolded pairs, the ground truth everything else is measured against.

Synthetic benchmark: `crates/chematic-fp/examples/superimposed_coding_spike_benchmark.rs`. Run it: `cargo run -p chematic-fp --release --example superimposed_coding_spike_benchmark`

## Provenance / COI (honest disclosure, as requested)

The general idea comes from a paper distributed under a **CC BY-NC** (non-commercial) license, whose authors disclosed a **conflict of interest** (a commercial fingerprint-tooling business, chemfp). **No code, pseudocode, or reported numbers from that paper were used anywhere in this PR.** Everything here is an independent design over the general *concept* of superimposed coding (a decades-old information-retrieval technique, Mooers 1949, popularized in bit-vector form by Bloom filters — not something the paper invented), verified only against synthetic data generated in this repo. Nothing here reproduces or claims to reproduce that paper's reported results.

## Structural fact, proved not just measured: `codeword_weight` cannot help

`codeword_weight=1` is **provably identical** to `count_simulation_fold` whenever `repetitions == max_repeats` and both use the same `seed` (true throughout this benchmark and the default config) — same `coded_bit(feature_hash, layer/unit, 0, seed, n_bits)` formula either way. Confirmed by both a unit test (`superimposed_with_codeword_weight_one_equals_count_simulation_fold`, using `repetitions=max_repeats=4`, matching seeds) and by the benchmark (their numbers match to 4 decimal places at every row below, same `repetitions=max_repeats=4` setup). That means `codeword_weight` is the *only* structural knob this spike's design adds over the existing RDKit-style baseline (given matched `repetitions`/`max_repeats`/`seed`). `SuperimposedCodingConfig::default()` now ships `codeword_weight: 1` — the proof below shows `>1` can only match or underperform `1`, never beat it, so the default shouldn't hand out the worst arm.

In the module docs I worked out why that knob can't help: in the collision-free limit (`n_bits → ∞`), a feature `f` with `L_f` active layers contributes exactly `codeword_weight` distinct bits per layer, so `|A| = codeword_weight · Σ_f L_f^A`, and two fingerprints only ever share a bit when both have the *same* layer active for the *same* feature (the hash is a pure function of its inputs), in which case all `codeword_weight` slots of that layer collide identically on both sides. So `|A∩B| = codeword_weight · Σ_f min(L_f^A, L_f^B)` and `|A∪B| = codeword_weight · Σ_f max(L_f^A, L_f^B)` — the `codeword_weight` factor cancels exactly in the Tanimoto ratio. On any finite `n_bits`, `codeword_weight > 1` only adds more chances for *unrelated* features to collide on the same bit — it can only hurt, never help. The benchmark's sweep (below) confirms this precisely: MAE rises and correlation falls monotonically as `codeword_weight` increases, tracking rising mean bit-vector fill.

## Measured numbers

Two regimes were run, because the first one I tried turned out to be misleading on its own (see the "what changed" note below).

**Regime A — realistic fill, clamp-free** (`n_features=60`, `max_count=4=max_repeats`, so no arm clamps counts and single-fingerprint popcount lands in real ECFP4's ballpark — tens of active features, not hundreds): pooled `overlap_frac ~ Uniform[0.05, 0.95)`, n=400 pairs, MAE = mean absolute error vs exact count-Tanimoto.

| n_bits | plain MAE | count_sim MAE (=`w=1`) | superimposed `w=2` MAE | `w=3` MAE | `w=5` MAE |
|---|---|---|---|---|---|
| 512 | 0.1958 | **0.0712** | 0.1386 | 0.2021 | 0.3174 |
| 1024 | 0.1854 | **0.0342** | 0.0684 | 0.1020 | 0.1664 |
| 2048 | 0.1793 | **0.0187** | 0.0367 | 0.0535 | 0.0878 |

Pearson r vs ground truth in this regime is ≥0.94 for every arm (plain and count_sim both ≥0.977), so MAE is the metric that differentiates them here, not correlation.

**Regime B — headline / high-fill** (`n_features=400`, `max_count=12 > max_repeats=4`, i.e. every count/count_sim/superimposed arm clamps): pooled results, plus the `ceiling(clamp@R)` row (`exact_count_tanimoto` on *both* fingerprints' counts clamped to `max_repeats`, zero hashing) that isolates how much of an arm's MAE is clamp bias vs. hashing/collision noise:

| n_bits | ceiling MAE | plain MAE | count_sim MAE | `w=2` | `w=3` | `w=5` |
|---|---|---|---|---|---|---|
| 512 | 0.1221 | 0.3365 | 0.5379 | 0.6723 | 0.7008 | 0.7073 |
| 1024 | 0.1213 | 0.2746 | 0.3711 | 0.5353 | 0.6267 | 0.6914 |
| 2048 | 0.1202 | 0.2415 | 0.2542 | 0.3711 | 0.4652 | 0.5917 |

At `n_bits=2048`, roughly half of `count_sim`'s 0.2542 MAE (0.1202) is clamp bias alone, not hashing noise — this regime's `max_count=12` vastly exceeds `repetitions=4`, so even a hypothetical zero-collision encoder couldn't do much better than the ceiling row. This regime also sits at 40–100% mean bit-vector fill (see the `mean_fill` column in the full benchmark output), far denser than real ECFP4 fingerprints (~50–100 active features into 2048 bits, ~5–10% fill) — it's a stress-test of saturation behavior, not a realistic operating point.

Both regimes agree on the one finding this spike set out to check: **`codeword_weight>1` never wins.** `w=1` (=`count_sim`) is the best of the superimposed-coding sweep in every row of both tables; `w=2/3/5` are monotonically worse.

## What changed from my first pass at this write-up (being upfront about it)

My first draft measured only Regime B and concluded "plain folding is competitive with or better than every hash-based strategy." That conclusion doesn't survive the more realistic Regime A, where `count_sim` clearly beats `plain` (e.g. 0.0187 vs 0.1793 MAE at `n_bits=2048`) — Regime B's apparent "plain wins" was an artifact of stacking two confounds (clamp bias from `max_count=12` vs `repetitions=4`, and saturation from `n_features=400` pushing every hash-based arm past 40% fill) that don't reflect how these fingerprints are actually sized in practice. Also worth being explicit: `gen_pair` gives shared features *independent* random counts in A and B; real similar molecules would have correlated counts, which is part of why ground truth sits well below presence-only Jaccard at high overlap in these synthetic pairs.

## Recommendation

**Do not productionize `codeword_weight>1`** — it's now both mathematically provable (above) and empirically confirmed in the realistic regime that it cannot beat plain `count_sim` (`codeword_weight=1`), and it never beats it in the unrealistic high-fill regime either. That part of this spike's naive generalization is a dead end, not a promising direction to iterate on further within this design.

Separately (and not really what this spike was scoped to answer): the existing RDKit-style `count_simulation_fold` baseline itself looks genuinely useful at realistic fill levels (MAE 3–10x lower than plain presence folding across `n_bits` in Regime A) — but that's an existing, well-known technique, not something this spike needed to establish, and productionizing *that* would be a separate, much larger decision (touching real molecule data, real fingerprint sizes, and the two-engine ECFP4 policy in `CLAUDE.md`) that's out of scope here. If "superimposed coding" specifically (i.e., something better than plain count-simulation) is worth revisiting, it would need a structurally different design than the one implemented here — e.g. keeping the bit budget closer to `O(log count)` via log-scale count-bound layers (`count ≥ {1,2,4,8,16}`) instead of unary-per-unit layers — which was not attempted in this bounded spike.

## Verification

- `cargo test -p chematic-fp --lib --quiet`: **267 passed** (250 pre-existing + 17 new in `superimposed_coding`), 0 failed.
- `cargo test -p chematic-fp --tests --quiet`: 267 (lib) + 2 (integration, incl. `rdkit_morgan_stable_api_fixtures` — the ECFP4 bit-exact guard) passed, 0 failed.
- `git diff --stat origin/main -- crates/chematic-fp`: only `src/lib.rs` changes (+7 lines, one new `pub mod` declaration + doc comment) plus the two new files — nothing under `ecfp.rs` / `rdkit_morgan_*.rs` / any existing file touched.
- `bash scripts/check.sh`: **all checks passed** (fmt, clippy `-D warnings`, full workspace lib + integration tests, cargo-deny, publish graph, version consistency).

## Explicitly out of scope (per the spike's own bounds, not attempted)

Wiring into `ecfp4()`/`mol.ecfp4()`/any public fingerprint API, adding to the RDKit-bit-exact engine, performance optimization, `chematic-py`/`chematic-wasm` bindings, and (per the recommendation above) the log-scale-layer redesign that might actually make `codeword_weight`-style superimposed coding work.

### DIFF
--- a/crates/chematic-fp/examples/superimposed_coding_spike_benchmark.rs
+++ b/crates/chematic-fp/examples/superimposed_coding_spike_benchmark.rs
@@ -19,14 +19,29 @@
 //! unit test) -- the *only* structural knob this spike's design adds over
 //! the RDKit-style baseline is `codeword_weight > 1` (multiple independently
 //! hashed bits per unary count-layer instead of one). The sweep exists to
-//! show, numerically, what that knob alone buys or costs.
+//! show, numerically, what that knob alone buys or costs. See
+//! `superimposed_coding`'s module docs for a short proof that
+//! `codeword_weight` is exactly Tanimoto-neutral in the collision-free
+//! limit, so any effect visible here is collision/saturation noise, never a
+//! genuine improvement.
 //!
-//! A separate pooled section at the end of each `n_bits` block draws
-//! `overlap_frac` uniformly at random per pair (instead of 5 fixed buckets)
-//! so ground truth varies continuously across the sample -- bucketed Pearson
-//! over 80 near-identical ground-truth values per bucket is dominated by
-//! noise, not signal; Pearson only means something once truth has real
-//! spread to correlate against.
+//! Every count/count_sim/superimposed arm clamps count at `max_repeats`, so
+//! their best-case output (even with zero hash collisions) is the exact
+//! Tanimoto of *clamped* counts, not the unclamped ground truth. A
+//! `ceiling(clamp@R)` row -- `exact_count_tanimoto` on both fingerprints'
+//! counts clamped to `max_repeats`, no hashing/folding at all -- is printed
+//! alongside every table so the clamp-bias contribution to each arm's MAE
+//! can be told apart from hashing/collision noise.
+//!
+//! Two regimes are covered per `n_bits`: bucketed fixed-overlap levels (MAE
+//! only -- Pearson over 80 near-identical ground-truth values per bucket is
+//! noise-dominated) and a pooled continuous-overlap sample (`overlap_frac ~
+//! Uniform[0.05, 0.95)` per pair, meaningful Pearson). A final section
+//! re-runs the pooled comparison in a smaller, clamp-free regime
+//! (`n_features=60`, `max_count=max_repeats`) closer to real ECFP4 fill
+//! levels (~5-10% of 2048 bits for drug-like molecules, vs. ~15-45% for the
+//! `n_features=400` headline numbers) to check whether the ordering
+//! observed at high, unrealistic fill survives at realistic fill.
 //!
 //! Usage:
 //! ```text
@@ -73,8 +88,11 @@ impl Rng {
 
 /// Generate a synthetic pair of count fingerprints sharing `overlap_frac` of
 /// a `n_features`-sized universe. Shared features get independent random
-/// counts in `A` and `B`; non-shared features are split evenly between the
-/// two so both fingerprints have private, non-overlapping content too.
+/// counts in `A` and `B` (a simplification: real similar molecules would
+/// have *correlated* counts on shared features, which would push ground
+/// truth higher at high overlap than this generator does); non-shared
+/// features are split evenly between the two so both fingerprints have
+/// private, non-overlapping content too.
 fn gen_pair(
     rng: &mut Rng,
     n_features: usize,
@@ -96,6 +114,13 @@ fn gen_pair(
         }
     }
     (a, b)
+}
+
+/// Clamp every count in `fp` to `cap` -- the best-case (zero-collision)
+/// output shape of any `count_sim`/`superimposed` arm with `max_repeats ==
+/// cap`.
+fn clamp_counts(fp: &[(u64, u32)], cap: u32) -> CountFp {
+    fp.iter().map(|&(f, c)| (f, c.min(cap))).collect()
 }
 
 fn pearson(xs: &[f64], ys: &[f64]) -> f64 {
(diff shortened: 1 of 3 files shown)


kent-tokyo:
PRの結論(Do not merge)通り、production化しない判断です。research findingは tasks/lessons.md に短く転記済み(PR本文もGitHub上に残ります)。unmerged closeとします。

PR #272, opened 2026-08-09
feat(smiles): expose topological_equivalence_classes

### LINKED ISSUE #265: chematic-smiles: expose canonical_automorphism/canonical_partition (topological symmetry classes)
## Summary

Expose the graph automorphism / canonical partition-refinement machinery that `chematic-smiles`'s canonical SMILES writer already uses internally, as a public API returning atom equivalence classes.

## Motivation

`chematic-smiles::canonical_automorphism` and `canonical_partition` (0.12.0) implement topological symmetry-class computation — needed internally to produce a canonical atom ordering — but neither module currently has a single `pub` item. There's no way for a downstream consumer to ask "which atoms are topologically equivalent to which other atoms" without reimplementing partition refinement from scratch, duplicating an algorithm chematic already has, tests, and relies on for its own canonicalization correctness.

This blocks things like meso-compound detection (a molecule with stereocenters that map onto each other under an internal mirror symmetry) and detecting "polyfunctional symmetry breaking" (whether a set of nominally-equivalent functional groups are actually related by a molecular symmetry, vs. only superficially similar).

## Proposed API

```rust
// Rust
use chematic_smiles::topological_equivalence_classes;
let classes: Vec<usize> = topological_equivalence_classes(&mol); // classes[i] = equivalence-class id of atom i
```

## RDKit / OpenBabel equivalent

RDKit's `Chem.CanonicalRankAtoms(mol, breakTies=False)` returns exactly this shape: ranks with ties preserved, i.e. equivalence classes.

## Additional context

Found while investigating meso-compound detection in a downstream consumer (kent-tokyo/yomitoki) — the algorithm already exists inside chematic, it just isn't reachable from outside the crate.

kent-tokyo:
## Summary

Fixes #265. Exposes the graph-automorphism / partition-refinement machinery `chematic-smiles`'s canonical-SMILES writer already uses internally (`canonical_partition.rs`'s `CanonicalColoredGraph` / `initial_partition` / `exact_refine`) as a new public function:

```rust
pub fn topological_equivalence_classes(mol: &Molecule) -> Vec<usize>
```

`classes[i] == classes[j]` iff atoms `i` and `j` are indistinguishable by exact structural color refinement (element, isotope, charge, aromaticity, effective H-count, bond order/connectivity). Re-exported at the crate root (`chematic_smiles::topological_equivalence_classes`); the `canonical_partition` module itself stays private, same as before.

## Mechanism (no parallel construction path)

`CanonicalColoredGraph` gains a `new_topological()` constructor next to the existing `new()` (still used unchanged by `canonical_search.rs` — zero call-site changes, zero risk to canonical-SMILES output). Both go through the same `build()`/`vertex_color()`/`edge_color()` helpers with one bool flag (`canonical_fidelity`) — no second, independent way to build the colored graph.

`topological_equivalence_classes` calls `new_topological`, then `exact_refine(initial_partition(graph, &vec![0; n]))`. The all-zero starting `ranks` collapses `initial_partition` to "group by vertex color alone" — the correct from-scratch starting point for a whole-molecule query (canonical-SMILES's own use seeds from a partially-individualized search-node state instead). This also means the result never depends on `crate::canonical::normalize_ranks`'s raw-FNV-1a-hash grouping — `exact_refine`'s own iterations only ever compare full signatures via real `Eq`/`Ord`.

## The canonicalization-vs-symmetry distinction (the issue's central ask)

`CanonicalColoredGraph::new`'s existing coloring folds in several devices that exist only to make canonical-SMILES search-pruning *safe*, not to describe real molecular symmetry. Applying any of them to a general symmetry query would silently turn "topological symmetry" into "canonical tie-break" — exactly the failure mode the issue called out for meso-compound detection. `new_topological` excludes all four, each verified empirically, not just reasoned about:

| Device | `new()` (canonical-SMILES) | `new_topological()` (this PR) | Evidence |
|---|---|---|---|
| Atom parity (`@`/`@@` + `stereo_unique`) | Forces every stereo atom (and neighbors) globally unique | Ignored | meso-tartaric acid `OC(=O)[C@H](O)[C@@H](O)C(=O)O`: `classes[3] == classes[5]` despite one atom being `CounterClockwise` and the other `Clockwise` — that opposite-tag pair *is* the meso relationship |
| H-count spelling | Raw bracket `hydrogen_count` (`[Cl]` ≠ organic-subset `Cl`) | Effective count (`implicit_hcount`) | `[cH]1ccccc1` (benzene, one H spelled explicitly) — reverting the fix reproduces `[3, 2, 1, 0, 1, 2]` (atom 0 isolated); fixed it returns all six atoms in one class |
| Reaction atom-map numbers | Included (writer must round-trip `[CH3:1]`) | Ignored | `[CH3:1]c1ccc(cc1)[CH3:2]`: measured *with* atom-map still folded in (pre-decision probe), the two chemically identical methyls split into different classes (`[0,2,4,5,3,5,4,1]`, indices 0 and 7 differ); after excluding it, they merge |
| E/Z bond direction (`Up`/`Down` order) | Distinct edge color from `Single` | Collapsed to `Single` | `F/C=C\F` (cis-1,2-difluoroethene, real mirror symmetry across the two `=CF` ends): before the fix, 4 singleton classes (`[2, 0, 1, 3]`); after, 2 merged pairs |

Each row has a regression test in `canonical_partition.rs`. `Molecule::bond_direction`'s separately-stashed direction (used for a ring bond next to an exocyclic stereo double bond) is not part of `EdgeColor` in either mode today, so there's nothing to exclude there — noted in the doc comment as a "if this ever gets wired in, gate it too" caution, not something requiring a fix or test now.

## Relationship to the existing `equivalent_atom_classes`

`chematic-smiles` already has a public, similarly-shaped function: `crate::canonical::equivalent_atom_classes` (built on `morgan_ranks`/`refine_ranks`), also chirality-blind (`initial_invariant` never reads `atom.chirality`). On all four structural test fixtures (benzene, toluene, an asymmetric chain, meso-tartaric acid), the two functions return **identical orbit partitions** (different numeric labels, same grouping) — measured, not assumed:

```
c1ccccc1:                       topological=[0,0,0,0,0,0]              equivalent_atom_classes=[0,0,0,0,0,0]
Cc1ccccc1:                      topological=[0,1,2,3,4,3,2]            equivalent_atom_classes=[1,2,0,3,4,3,0]
FCCCCCl:                        topological=[4,2,0,1,3,5]              equivalent_atom_classes=[4,3,2,0,5,1]
OC(=O)[C@H](O)[C@@H](O)C(=O)O:  topological=[3,0,2,1,4,1,4,0,2,3]      equivalent_atom_classes=[1,2,0,3,4,3,4,2,0,1]
```

So this PR is not "fixing" a wrong answer from `equivalent_atom_classes` on any case tested. It is a narrower justification than that, and worth stating plainly rather than overselling: two real reasons to still ship this as a separate function:

1. The issue specifically asks for `canonical_automorphism`/`canonical_partition` — the exact-refinement machinery — exposed, not for wrapping the existing hash-based path.
2. `exact_refine` never merges two cells on anything but full signature equality; `equivalent_atom_classes`'s `normalize_ranks` groups by raw FNV-1a hash-value equality (a real, if practically negligible, collision risk — see this module's own top-of-file doc comment).

(`atom_map` is *not* a differentiator, despite being the one place these two functions could plausibly disagree: `initial_invariant` already never reads `atom.atom_map`, and this PR's `vertex_color` independently landed on excluding it too — see the table above for the measurement that drove that decision. The two functions agree here by two independent routes to the same conclusion, not by coincidence.)

Side note, out of scope for this PR: `equivalent_atom_classes`'s own doc comment gives a stale example (`Cc1ccccc1` → `[0,1,1,1,1,1,2]`, i.e. 3 classes). Measured actual output is `[1,2,0,3,4,3,0]`, i.e. **5** classes (methyl / ipso / ortho-pair / meta-pair / para) — matching this PR's own toluene test. Not fixing the stale doc here; flagging it since this PR's investigation surfaced it.

## Known limitation (shared with `morgan_ranks`/`equivalent_atom_classes`)

Like any Morgan/1-WL-style color refinement, `exact_refine` is a sound *over-approximation* of true automorphism orbits: same class does not strictly prove "provably interchangeable by a real automorphism" (only the converse holds). On a pathological WL-indistinguishable disjoint union of differently-sized regular subgraphs it can under-split — see `canonical_automorphism::tests::triangle_and_square_same_wl_cell_different_orbits` for a constructed witness. No real molecule in this crate's test corpus has ever exhibited this; documented, not fixed, exactly as it already is for `morgan_ranks`.

## Test coverage

10 new tests in `canonical_partition.rs`:
- `benzene_all_six_carbons_one_class` — obvious symmetry, all 6 in one class.
- `toluene_five_classes_ortho_and_meta_pairs_merge` — **5** distinct classes (methyl, ipso, ortho-pair, meta-pair, para), reasoned through and verified, not asserted from the issue text's own (self-questioning) number.
- `asymmetric_chain_every_atom_its_own_class` (`FCCCCCl`) — 6 distinct classes, no duplicates.
- `meso_tartaric_acid_stereocenters_share_a_class_despite_opposite_tags` — the issue's stated real-world motivation.
- `bracket_h_spelling_does_not_split_the_class`, `atom_map_number_alone_does_not_split_the_class`, `ez_bond_direction_marker_alone_does_not_split_the_class` — regression tests for the three divergences found during review.
- `wildcard_never_collides_with_a_zero_h_real_atom` — confirms dropping the canonical branch's H-count offset didn't introduce a wildcard/real-atom collision.
- Plus the two pre-existing `canonical_partition` tests, unchanged.

Full crate suite: 202 tests, `cargo clippy --all-targets -D warnings` clean, `cargo doc` clean (no `private_intra_doc_links`), `cargo fmt --check` clean, and `bash scripts/check.sh` (full workspace: fmt, clippy, all lib + integration tests, cargo-deny, publish graph, version consistency) passes clean.

## Out of scope

- Python (`chematic-py`) / WASM (`chematic-wasm`) bindings — the issue's proposed API is Rust-only; `equivalent_atom_classes` already has a `mol.equivalent_atom_classes()` Python binding as a working precedent if a follow-up wants one for this function too.
- Fixing `equivalent_atom_classes`'s stale doc example (noted above).

## Test plan

- [x] `cargo test -p chematic-smiles --lib` (202/202)
- [x] `cargo clippy -p chematic-smiles --all-targets -- -D warnings`
- [x] `cargo doc -p chematic-smiles --no-deps` (no warnings)
- [x] `cargo fmt --all -- --check`
- [x] `bash scripts/check.sh` (full workspace, final tree including the E/Z fix)

### DIFF
--- a/crates/chematic-smiles/src/canonical_partition.rs
+++ b/crates/chematic-smiles/src/canonical_partition.rs
@@ -547,7 +547,12 @@ pub(crate) fn exact_refine(graph: &CanonicalColoredGraph, mut partition: Partiti
 ///   difluoroethene, real mirror symmetry across the two `=CF` ends) came
 ///   back as 4 singleton classes before this collapse was added, 2 merged
 ///   pairs after (see `ez_bond_direction_marker_alone_does_not_split_the_class`
-///   below).
+///   below). Note `edge_color` only ever reads `bond.order` -- the
+///   separately-stashed `Molecule::bond_direction` (`bond_has_direction_info`'s
+///   doc comment: used for a ring bond next to an exocyclic stereo double
+///   bond) is not part of `EdgeColor` in either mode, so there is nothing
+///   to exclude there today; if a future change ever folds it in, it needs
+///   the same `canonical_fidelity` gate.
 ///
 /// `CanonicalColoredGraph::new_topological` therefore omits all four,
 /// using the *effective* H count (`implicit_hcount`, matching
--- a/crates/chematic-smiles/src/canonical_partition.rs
+++ b/crates/chematic-smiles/src/canonical_partition.rs
@@ -240,6 +240,17 @@ fn vertex_color(
 pub(crate) struct CanonicalColoredGraph<'a> {
     mol: &'a Molecule,
     vcolor: Vec<VertexColor>,
+    /// See `new`/`new_topological`. Also gates `edge_color`: an `Up`/`Down`
+    /// bond order is itself an E/Z *direction* marker on what is
+    /// chemically a single bond (`bond_has_direction_info`'s doc comment),
+    /// so it is exactly as canonicalization-only as the vertex-side stereo
+    /// devices `vertex_color` documents -- omitted here for the same
+    /// reason. Caught empirically, not by inspection alone: an earlier
+    /// version of this PR left `edge_color` reading raw `Up`/`Down`
+    /// unconditionally, and `F/C=C\F` (cis-1,2-difluoroethene, a real
+    /// mirror-symmetric molecule -- swap the two `=CF` ends) came back as
+    /// 4 distinct singleton classes instead of the correct 2 merged pairs.
+    canonical_fidelity: bool,
 }
 
 impl<'a> CanonicalColoredGraph<'a> {
@@ -253,15 +264,17 @@ impl<'a> CanonicalColoredGraph<'a> {
     }
 
     /// Coloring for [`crate::topological_equivalence_classes`]: identical to
-    /// `new` except every canonicalization-only device `vertex_color`
-    /// documents is turned off (see `topological_equivalence_classes`'s doc
-    /// comment for the full reasoning -- the short version: `new`'s stereo
-    /// and raw-H-spelling handling exist to make canonical-SMILES
-    /// tie-breaking conservative/safe, a different goal from reporting real
-    /// topological symmetry; applying either here would wrongly split
-    /// atoms that a real symmetry query must merge -- a meso compound's two
-    /// mirror-equivalent stereocenters via stereo, or `[Cl]` vs `Cl` via raw
-    /// H-spelling).
+    /// `new` except every canonicalization-only device `vertex_color` (and
+    /// this struct's own `edge_color`) documents is turned off (see
+    /// `topological_equivalence_classes`'s doc comment for the full
+    /// reasoning -- the short version: `new`'s stereo and raw-H-spelling
+    /// handling exist to make canonical-SMILES tie-breaking
+    /// conservative/safe, a different goal from reporting real topological
+    /// symmetry; applying any of them here would wrongly split atoms that a
+    /// real symmetry query must merge -- a meso compound's two
+    /// mirror-equivalent stereocenters via stereo, `[Cl]` vs `Cl` via raw
+    /// H-spelling, or a cis-alkene's two mirror-equivalent ends via
+    /// `Up`/`Down` bond-direction markers).
     pub(crate) fn new_topological(mol: &'a Molecule) -> Self {
         Self::build(mol, false)
     }
@@ -278,7 +291,11 @@ impl<'a> CanonicalColoredGraph<'a> {
                 vertex_color(mol, idx, sensitive.contains(&idx), canonical_fidelity)
             })
             .collect();
-        Self { mol, vcolor }
+        Self {
+            mol,
+            vcolor,
+            canonical_fidelity,
+        }
     }
 
     pub(crate) fn mol(&self) -> &'a Molecule {
@@ -296,8 +313,17 @@ impl<'a> CanonicalColoredGraph<'a> {
     pub(crate) fn edge_color(&self, from: AtomIdx, bidx: BondIdx) -> EdgeColor {
         let bond = self.mol.bond(bidx);
         let from_is_donor = bond.order == BondOrder::Dative && bond.atom1 == from;
+        // `Up`/`Down` are E/Z direction markers on an otherwise-single
+        // bond; in topological mode they collapse to plain `Single`, same
+        // as every other stereo device this struct excludes there.
+        let order =
+            if !self.canonical_fidelity && matches!(bond.order, BondOrder::Up | BondOrder::Down) {
+                BondOrder::Single
+            } else {
+                bond.order
+            };
         EdgeColor {
-            order_class: order_class(bond.order),
+            order_class: order_class(order),
             from_is_donor,
         }
     }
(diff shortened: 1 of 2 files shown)


kent-tokyo:
Status update (2026-08-11): same verdict as before — not merging as-is — but now with a concrete destination.

**Why not merge as-is**: `topological_equivalence_classes` is currently a 1-WL (color-refinement) over-approximation published under a name that implies more than it actually proves — 1-WL can under-refine (merge atoms that aren't truly topologically equivalent) for some symmetric/regular graph structures, so the name promises exactness the implementation doesn't deliver.

**Where it's going**: this feeds v0.14.0's A-tier item 3, the same shared bounded/typed exact-automorphism-orbit kernel (`exact_orbits()`, `{Exact, LimitExceeded}`) being built for #257, meso-compound detection, #139's VF2 pruning, and the new conformer-ensemble dedup step. `topological_equivalence_classes` becomes a consumer of that shared kernel rather than its own independent (and currently over-approximating) engine — this closes the standing objection to merging #257 and #272 as two divergent symmetry code paths.

**Design input worth reading first** (ROADMAP.md, added 2026-08-10): *StereoMolGraph* (ChemRxiv 2025-12-28, JCIM 2026-03-27) argues for treating topological equivalence and stereochemical equivalence as genuinely separate notions, with local stereo descriptors handled as their own thing rather than a parity tie-break bolted onto the color-refinement pass — the same shape of gap this PR's current 1-WL-only approach has on the topology axis.

Keeping this PR open, tracked as an input to the shared-engine redesign, not closing it. No action needed from here for now.

PR #280, opened 2026-08-10
fix(perception): fix implicit-H rank-0 sentinel collision in stereo_completeness

### LINKED ISSUE #267: chematic-perception: stereo_completeness/validate_stereo panic (debug) or silently corrupt (release) on any negatively-charged atom
## Description

`chematic_perception::stereo_validation::simple_morgan_ranks` (private helper used by both `stereo_completeness` and `validate_stereo`) computes each atom's initial invariant as:

```rust
atom.element.atomic_number() as u64 * 1_000_000 + atom.charge as u64 * 1000 + deg
```

(`stereo_validation.rs:82`). `Atom.charge` is `i8`. For any atom with a **negative** formal charge, `atom.charge as u64` sign-extends before reinterpreting as unsigned (e.g. `-1i8 as u64 == u64::MAX`), so `atom.charge as u64 * 1000` overflows `u64` unconditionally, regardless of the actual charge magnitude.

- **Debug builds**: this overflow check panics (`attempt to multiply with overflow`), taking down any caller. Since `stereo_completeness` and `validate_stereo` are both meant to be safe, infallible perception functions, this is a crash on ordinary, common input — any carboxylate, sulfonate, phosphate, or other anion.
- **Release builds**: no panic (wrapping arithmetic), but the invariant computation is now nonsense for that atom, silently corrupting the Morgan-rank refinement `stereo_completeness`/`validate_stereo` are built on. In the two cases we checked directly, the final stereocenter count still came out correct (the corrupted value happened to stay distinct enough for the tie-break logic) — but that's incidental, not something the algorithm's design guarantees, since the entire point of the invariant is to be a small, meaningfully-ordered per-atom number, and an overflow-wrapped `u64::MAX`-derived value defeats that.

## Reproduction

```rust
use chematic_smiles::parse;
use chematic_perception::stereo_validation::stereo_completeness;

let acetate = parse("CC(=O)[O-]").unwrap();
stereo_completeness(&acetate); // debug: panics ("attempt to multiply with overflow")
                                // release: returns Ok-looking StereoCompleteness, but via
                                // wrapped-overflow arithmetic, not the intended computation

let cation = parse("C[NH3+]").unwrap();
stereo_completeness(&cation);  // fine either way — small positive i8 charge, no overflow

let salt = parse("CC(=O)[O-].C[NH3+]").unwrap();
stereo_completeness(&salt);    // debug: panics (same as acetate alone)
```

Minimal debug-mode repro (`cargo test`, default profile):

```rust
#[test]
fn repro() {
    let mol = chematic_smiles::parse("CC(=O)[O-]").unwrap();
    let _ = chematic_perception::stereo_validation::stereo_completeness(&mol); // panics here
}
```

## Expected behavior

`stereo_completeness`/`validate_stereo` should handle negatively-charged atoms the same way they already handle positively-charged ones: correctly, without overflow, in both debug and release profiles. A formal charge in the realistic range for any element on the periodic table (roughly -4..+8) should never come close to overflowing a `u64` component of this invariant regardless of sign.

## Actual behavior

Any atom with `charge < 0` triggers a guaranteed arithmetic overflow in the initial-invariant computation — panics in debug builds, silently produces a corrupted invariant in release builds.

## Environment

- chematic version: 0.12.0 (via the `chematic` facade crate, `perception` feature)
- OS: macOS

---

Found via `kent-tokyo/yomitoki`, which calls `stereo_completeness` unconditionally on every `analyze`/`analyze_smiles` call (both in its `applicability` and `stereochemical_burden` components) — meaning this currently panics on *any* input molecule containing a negatively charged atom (any carboxylate/sulfonate/phosphate anion, extremely common in real chemistry) when built in debug mode, which is the default `cargo build`/`cargo test` profile. We're adding a guard on our side to avoid calling into this until it's fixed, but wanted to flag the root cause rather than just work around it silently.

kent-tokyo:
## Root cause

`crates/chematic-perception/src/stereo_validation.rs`'s `simple_morgan_ranks` normalises each atom's Morgan-refinement invariant to a consecutive ordinal starting at **0** (its tail: `sorted.partition_point(|&u| u < *r)`). `stereo_completeness` used the literal value `0` as a stand-in "rank" for an implicit hydrogen when checking whether a candidate stereocenter's 4 groups (heavy-atom neighbours + optionally 1 implicit H) are all distinct:

```rust
let mut nb_ranks: Vec<u64> = heavy_nbs.iter().map(|nb| ranks[nb.0 as usize]).collect();
if implicit_h > 0 {
    nb_ranks.push(0);
}
...
if sorted.len() < 4 { continue; } // symmetric neighbours — not a stereocenter
```

Rank `0` is **not** reserved — it's an ordinary, frequently-reached normalised rank (whichever atom in the whole molecule has the lowest invariant gets it). When a candidate stereocenter has a heavy neighbour that carries real rank `0` *and* also has an implicit H, `nb_ranks` ends up with two entries equal to `0`; `dedup()` collapses them, `sorted.len()` drops below 4, and the atom is silently treated as "not a stereocenter" even though it has 4 genuinely distinct groups and may carry an explicit `@`/`@@` annotation. This **undercounts** `stereo_completeness`'s `specified` count.

This is a distinct bug from #267 (a `u64`-overflow/sign-extension bug in the same function, fixed in #279) — a rank-0 sentinel collision, not a magnitude issue. In fact #279's own PR body independently surfaced this exact mechanism as an unverified "leading explanation" for a ~10% diff it found on a 5,000-molecule corpus when it prototyped an alternate overflow fix, and explicitly deferred it as out of scope. This PR confirms the collision mechanism exists and undercounts `specified` (repro below). I did not re-run that 5,000-molecule diff, so this doesn't independently confirm the collision fully accounts for #279's observed delta — only that the mechanism it flagged is real.

## Repro (added as a regression test)

```
C[C@@H](Cl)C(Br)(F)I
```

Atom 1 is the `@@`-annotated chiral carbon. Its methyl neighbour (atom 0) is the lowest-invariant atom in the whole molecule (`simple_morgan_ranks` gives `[0, 6, 2, 5, 4, 1, 3]`), colliding with the implicit-H sentinel at atom 1.

- **Before fix:** `stereo_completeness(...) == StereoCompleteness { specified: 0, unspecified: 1, total_centers: 1 }` — the annotated stereocenter is dropped entirely; only the unrelated unspecified center at the quaternary carbon (atom 3) is counted.
- **After fix:** `StereoCompleteness { specified: 1, unspecified: 1, total_centers: 2 }` — both real stereocenters are counted.

A second regression test (`OC[C@@H](N)C(Br)(F)I`) exercises the same collision class with a structurally different colliding neighbour (a substituted `-CH2-OH` carbon rather than a bare terminal methyl), confirmed via the same `simple_morgan_ranks` debug print (`ranks = [1, 0, 5, 3, 7, 6, 2, 4]`, atom 1 = the CH2-OH carbon), to rule out the fix being lucky on one geometry.

## Fix

Replace the literal `0` sentinel with `mol.atom_count() as u64`, which is provably never a reachable real rank: the maximum normalised rank from `simple_morgan_ranks` is `(number of distinct invariant values) - 1 <= atom_count() - 1`.

## `validate_stereo` (checked, not changed)

`validate_stereo`'s Rule 3 (`if all_same && implicit_h == 0`) uses `implicit_h` as a plain hydrogen **count**, never pushed into a rank vector alongside real ranks — confirmed by reading the code, not just assumed. It is not affected by this bug and was left unchanged.

## Testing

- Two new regression tests in `stereo_validation.rs`'s test module (`test_stereo_completeness_rank_zero_collision_chain`, `test_stereo_completeness_rank_zero_collision_different_neighbor`). Verified by a before/after diff on the same two molecules (stash the fix, re-measure `stereo_completeness` directly): pre-fix both report `specified: 0` (matching the bug description), post-fix both report `specified: 1` -- then the tests were written to assert the post-fix value.
- Full existing `chematic-perception` test module (11 tests, including all of #279's overflow regression tests) still passes unchanged — no previously-passing test's expected values shifted.
- `bash scripts/check.sh` passes in full: fmt, clippy `-D warnings`, full workspace lib + integration tests, cargo-deny, publish graph, and version consistency all green.
- No public API changes: `stereo_completeness`/`validate_stereo` signatures are untouched.

## Relation to other work

This is the **second** of two prerequisite fixes the repo owner specified before #268 ("expose per-atom stereocenter candidates") can be rebased and merged: fix #267 (#279) → fix implicit-H sentinel collision (this PR) → rebase #268 → add regression fixtures → CI green → merge. This PR does not touch #268 or its branch.

**Based on #279's branch** (`fix/simple-morgan-ranks-negative-charge-overflow-267`), not `main`, since both fixes touch the same function and this avoids a merge conflict. **This PR must be re-targeted to `main` (or merged after) once #279 lands** — right now it is stacked on top of it.

---

**Not merging this myself.** Opening ready for review and stopping here.

### DIFF
--- a/crates/chematic-perception/src/stereo_validation.rs
+++ b/crates/chematic-perception/src/stereo_validation.rs
@@ -252,10 +252,23 @@ pub fn stereo_completeness(mol: &Molecule) -> StereoCompleteness {
             continue;
         } // only tetrahedral candidates
 
-        // Check all neighbours have distinct ranks (including implicit H as rank 0).
+        // Check all neighbours have distinct ranks (including implicit H as a
+        // sentinel rank). `simple_morgan_ranks` normalises to consecutive
+        // ordinals starting at 0 (see its tail: `partition_point(|&u| u <
+        // *r)`), so a real heavy-atom neighbour can legitimately carry rank
+        // 0 -- it's the ordinary "lowest invariant in the molecule" rank,
+        // not a reserved value. Using the literal `0` as the implicit-H
+        // sentinel therefore collided with real rank-0 neighbours (issue
+        // #267's follow-up bug): `dedup()` merged the two, `sorted.len()`
+        // dropped below 4, and a genuine 4-distinct-group stereocenter was
+        // silently skipped. The maximum normalised rank is (number of
+        // distinct invariants - 1), which is always <= atom_count() - 1, so
+        // `atom_count()` itself is never a reachable real rank and is safe
+        // to use as the sentinel here.
+        let implicit_h_rank_sentinel = mol.atom_count() as u64;
         let mut nb_ranks: Vec<u64> = heavy_nbs.iter().map(|nb| ranks[nb.0 as usize]).collect();
         if implicit_h > 0 {
-            nb_ranks.push(0);
+            nb_ranks.push(implicit_h_rank_sentinel);
         }
 
         let mut sorted = nb_ranks.clone();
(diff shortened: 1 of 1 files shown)


kent-tokyo:
Retargeted to main after #279 merged — reopening to trigger CI against the new base.

PR #315, opened 2026-08-13
feat(ff): port RDKit's eqLevel atom-type-equivalence ladder for Angle lookup (issue #227 Stage B)

kent-tokyo:
## Summary (Stage B of the 3-stage plan started in #314; stacked on top of it)

Stage A (#314) closed the Angle/Bond `type_mismatch` bucket. This PR closes the `equivalence_table` bucket by porting the mechanism RDKit's real Angle lookup actually has: `MMFFAngleCollection::operator()`'s 4-stage `eqLevel` canonical-type-substitution ladder (`Code/ForceField/MMFF/Params.h` at the pinned commit) — `angle_type` and the central atom type stay fixed; `type_i`/`type_k` are substituted through RDKit's real Level 2/3/4/5 equivalence classes before giving up. chematic had zero equivalent of this mechanism.

**Bigger effect than the original diagnostic predicted** (~9% of the *table_gap-only* residual): since `eqLevel` is an independent axis from atom-type routing correctness, it also incidentally rescues many `routing_bug_candidate` instances that weren't part of the originally-diagnosed 97-instance `table_gap` population at all.

## Changes

- `eq_level`/`MMFF94_EQ_LEVEL` (55-entry table, extracted from `defaultMMFFDef`) in `crates/chematic-ff/src/mmff94_energy/angle.rs`, inserted into `mmff94_angle_energy`'s search chain between the exact-match search and chematic's own pre-existing (non-RDKit) `angle_type=0` safety-net fallback.
- Updated one pre-existing test (`angle_type_for_butadiene_sp2_single_bond_...`) whose assertion "`angle_type=0` is a clean miss for this triple" no longer holds now that `eqLevel` makes it resolve too — to a genuinely *different*, wrong value (118.043° vs the correct 121.55°). Rewrote the assertion to demonstrate the more precise point: getting `angle_type` wrong now silently returns a *different wrong* parameter, not just a miss — the same silent-wrong-parameter failure class as the #236 furan collision.
- 2 new unit tests for `eq_level`/the ladder itself, oracle-verified.

## Verification

- Re-measured on the 265-molecule Wave 1 corpus: `mmff94_strict`'s bond+angle coverage gate moves **103 → 84 failing molecules** (19 more flip to pass beyond Stage A's own 4; **zero** regressions), `angles_missing` 358 → 191.
- `cargo test --workspace` green (all crates), including the updated/new tests.
- `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all -- --check` clean.
- `validation/results/mmff94_coverage_227_term_audit.jsonl` refreshed.
- Re-running `scripts/mmff94_angle_bond_gap_classify.py` post-fix: the Angle/Bond `table_gap` residual collapses to 8 + 1 unique tuples (from 27 + 3), **100% `empirical_rule`** — Stage C (eq.18-20, Halgren MMFF.V) is now the only remaining mechanism gap under issue #227, and 7/8 of its unique tuples are already oracle-confirmed by this same diagnostic.

**Not merging without a separate, explicit go-ahead**, per this repo's standing convention. This PR is stacked on #314 — merge #314 first (or merge both together), otherwise this PR's diff will include #314's changes too.


kent-tokyo:
Retargeting base to main and force-triggering CI (retarget alone doesn't fire the pull_request workflow trigger without a synchronize/reopened event).

PR #317, opened 2026-08-13
feat(ff): Halgren MMFF.V eq.18-20 empirical Bond/Angle rule (issue #227 Stage C2)

kent-tokyo:
Stacked on #316 (Stage C1: wildcard theta0 table restoration).

## Summary

- Ports RDKit's real `getMMFFBondStretchEmpiricalRuleParams`/`getMMFFAngleBendEmpiricalRuleParams` (Halgren MMFF.V eq. 18-20) as `bond_empirical`/`mmff94_bond_energy_resolved` and `angle_empirical_theta0`/`angle_empirical_ka`/`mmff94_angle_energy_resolved` — strictly *after* the existing exact-table/eqLevel-ladder chain. The empirical rule never overrides a real table hit.
- Adds `Mmff94Resolution` (`DirectTable`/`EquivalentType`/`GenericAngleTypeFallback`/`EmpiricalBond`/`EmpiricalAngle`) so a lookup's resolution mechanism is distinguishable — a table hit alone doesn't prove correctness (Stage B's `eqLevel` ladder can land on a real row that's the wrong parameter, same failure class as the #236 furan collision).
- New source tables, verbatim from the pinned `Params.cpp`: `MMFF94_COV_RAD_PAU_ELE`/`MMFF94_BNDK`/`MMFF94_HERSCHBACH_LAURIE` (Bond), `MMFF94_ANGLE_Z`/`MMFF94_ANGLE_C` (Angle).
- Ports `isAngleInRingOfSize3or4` verbatim (`is_angle_in_ring_of_size_3_or_4`) — confirmed by direct source read that the ring-size theta0 override applies only in the from-scratch branch, never in the `ka==0.0`-row theta0-reuse branch (#316), and preserves that asymmetry exactly.
- Wires the `_resolved` functions into production `bond_energy`/`angle_energy`/`stretch_bend_energy` and chematic-3d's `compute_mmff94_coverage`, so the coverage gate and the actual minimizer stay consistent.

## The one fail-closed tuple

`(angle_type=0, type_i=43, type_j=18, type_k=63)` (`chembl_tier_b_0022`) is deliberately **not** resolved. Numeric type 63 has no `MMFF94_EQ_LEVEL` row; RDKit's real eqLevel loop dereferences that unchecked (undefined behavior in RDKit's own C++, confirmed by direct source read). The live oracle's value is stable across 20 atom renumberings and a second unrelated molecule, but that's consistent with either a real resolution path or RDKit deterministically reading its own out-of-bounds memory — no source-level evidence resolves the ambiguity. Left fail-closed rather than guessed (see PROVENANCE.md for the full investigation).

## End-to-end confirmation

Two pre-existing `chematic-3d` tests asserted the *old* "no MMFF94 table entry, must fall back to UFF" behavior for `[C@H](F)(Cl)Br`'s 3 halogen-C-halogen angles (3 of the 7 oracle-confirmed empirical tuples). Both are rewritten to assert the new correct behavior — independent confirmation (a second code path, not just the new unit tests) that the empirical rule is genuinely wired into production.

## Corpus re-measurement

Full 265-molecule corpus, production `minimize_with_policy`, from the post-#314/#315 baseline (178 Ok / 83 MissingParameters / 1 UnsupportedAtomType / 3 MinimizationFailed = 87 failing):

- **This PR: 248 Ok / 13 MissingParameters / 1 UnsupportedAtomType / 3 MinimizationFailed → 17 failing (87 → 17).**
- **Zero regressions**, verified by full per-molecule join, not aggregate counts.
- `MinimizationFailed` set (`chembl_tier_b_0028`/`0029`/`0030`) and `UnsupportedAtomType` set are byte-identical before/after — not newly caused, not masked.
- All 70 `Ok`-transitions are `MissingParameters → Ok`.
- Of the 13 still `MissingParameters`: 12 have real, unrelated `Exhausted`-path Angle gaps out of this PR's scope; 1 is the fail-closed `chembl_tier_b_0022` above.
- Reproduced twice: once on a combined pre-split implementation, once independently after the C1/C2 split — **0 mismatches** between the two, confirming the split preserved exact functional equivalence.

## Test plan

- [x] `cargo test --workspace` green (verified on both C1-only and C1+C2 states)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo fmt --all -- --check` clean
- [x] New oracle-verified unit tests for both empirical paths (Bond + 7 Angle tuples), plus a fail-closed regression test for the 8th
- [x] 265-molecule corpus re-measurement: 87 → 17 failing, 0 regressions (see above)


kent-tokyo:
Retargeting base to main and force-triggering CI (retarget alone doesn't fire the pull_request workflow trigger without a synchronize/reopened event).

PR #326, opened 2026-08-15
Generalized stereo-configuration geometry (Tetrahedral + Square-planar)

kent-tokyo:
## Summary

Generalized stereo-configuration geometry foundation covering **Tetrahedral** and **Square-planar** (`chematic-core/src/stereo_geometry.rs`): a coordination geometry + the equivalence class of ligand-slot permutations under that geometry's *proper rotation group*. Replaces two previously independent, hand-written stereo-remapping algorithms in `chematic-smiles/src/canonical.rs` (`permutation_is_odd`, `remap_square_planar`) with one shared group-theoretic primitive, and fixes a real bug found along the way: a `SquarePlanar`-tagged atom could be silently coerced into a tetrahedral chiral-volume check in `chematic-3d`.

- Design doc: `docs/rfcs/generalized_stereo_geometry_rfc.md`
- Zero `vita` dependency anywhere (checked repo-wide across every file this PR touches, not just `Cargo.toml`s)
- No `Cargo.toml` changes, no `unsafe`, `#[non_exhaustive]` on `StereoGeometry` for future TBP/octahedral extension (not implemented here)

**Verified head: `cd24b27`** (7 commits). This PR went through one round of independent review after the initial implementation; see "Review round 2" below for what changed and why.

## Rotation-group derivation (the crux of the design)

- **Tetrahedral proper rotation group = A4, order 12** (all even permutations of the 4 ligand slots — a rigid tetrahedron's rotations permute its vertices by an even permutation only; an odd permutation requires a reflection). 24 orderings / 12 rotations = 2 orbits, matching `@`/`@@`.
- **Square-planar proper rotation group = order 8** (the S4-stabilizer of one trans-pair partition, e.g. SP1's `{0,2}|{1,3}`), **NOT** the naive "4 in-plane rotations only" group (order 4). The naive group would give 24/4 = 6 orbits and cannot recover SP1/SP2/SP3. Derived via orbit-stabilizer (S4 acts transitively on the 3 trans-pair partitions, orbit size 3, so stabilizer order = 24/3 = 8) and explicitly enumerated: 4 block-preserving elements (identity, (02), (13), (02)(13)) + 4 block-swapping elements ((01)(23), (0123), (0321), (03)(12)).
- **Both orders verified exhaustively, not just asserted**: group closure/identity/inverses/no-duplicate-rows tests for both tables; an independent from-scratch brute-force parity cross-check for the tetrahedral table (distinct code from `remap_tetrahedral_parity`'s own internals); and the flagship test enumerating **all 24** permutations of 4 distinct ligands (not sampled — plain nested loops generating the full permutation group) and confirming **exactly 2 distinct canonical representatives for Tetrahedral, exactly 3 for Square-planar**. A load-bearing test (`square_planar_rotations_stabilize_sp1_partition`) additionally ties the square-planar table to `SquarePlanarPermutation::SP1.trans_pairs()` *at runtime*, not a hardcoded partition literal — this is what would catch accidentally building the stabilizer of a *different* one of S4's 3 order-8 partition-stabilizers (a mistake the group-axiom and orbit-count tests alone cannot distinguish from the correct table, since all three give equally-valid, equally-3-orbit groups that just don't match `trans_pairs()`'s actual semantics).

## Existing tag semantics fully preserved

`@`/`@@`/`@SP1`/`@SP2`/`@SP3` mean exactly what they meant before this PR. Evidence:
- 88 stereo-bearing fixture SMILES (every `@`-containing literal already in `canonical.rs`/`canonical_robustness.rs`/`square_planar_stereo.rs`, plus all 18 SMILES in `validation/platinum/pt_corpus.jsonl`, plus 2 synthetic isotope/atom-map cases) produce **byte-identical canonical SMILES** before vs. after this PR's full diff (throwaway dump-and-diff tool, not committed), re-confirmed at the final head after review round 2's changes.
- That check **found and fixed a real transient regression during development**, not a residual issue left in the shipped PR: allene *end* carbons (e.g. `F[C@@H]=[C]=[C@H]Cl`) have a 3-element `stereo_neighbor_order` (sp2, one real double-bond partner standing in for the 4th position), not 4. `StereoGeometry::Tetrahedral` is fixed at 4 slots by design, so routing this case through it would silently pick the wrong fallback reason. Fixed by keeping the original, unmodified, length-generic `permutation_is_odd` as an explicit fallback whenever either side isn't exactly 4 elements — exact pre-existing behavior for this case, not a new algorithm. Now **pinned with an exact golden-value test** (`allene_stereo_exact_canonical_value_is_pinned`), verified to fail if the fallback is disabled (not just described) — the two pre-existing allene tests (`_two_enantiomers_differ`/`_round_trip_stable`) kept passing throughout the regression, since neither checks an exact value, only relative invariants.
- `remap_square_planar_tag` is provably equivalent to the removed `remap_square_planar`, not just observed to agree: a `SquarePlanar` rotation orbit is exactly the set of orderings sharing the same unordered trans-pair-of-ids partition, which is precisely what the old hand-written algorithm computed. All 8 pre-existing `square_planar_stereo.rs` tests (including `cisplatin_and_transplatin_have_distinct_canonical_identity`) pass unchanged, plus a **new** direct unit-level test (`remap_square_planar_tag_matches_oracle_table_directly`) exercising all 144 oracle cases against the new function directly, not just through the parser/writer.
- Cisplatin/transplatin-shaped **duplicate-chemistry** test (2×Cl + 2×NH3, distinct raw ids): `duplicate_chemistry_distinct_ids_keeps_cisplatin_transplatin_shaped_tags_distinct` proves SP1/SP2/SP3 never collapse into one orbit for chemically-repeated-but-id-distinct ligands, at the geometry-module level (independent of the pre-existing end-to-end test with the same property). The test's own comment initially misstated which of SP1/SP2 was cis- vs. trans-shaped for this specific fixture's neighbor order (self-contradictory: claimed "cis" while describing a trans arrangement) — fixed in review round 2 by dropping the specific cis/trans chemistry claim entirely rather than risk a second wrong derivation; the test now refers to "SP1-shaped"/"SP2-shaped" configurations. This was a comment-only fix — the assertion logic (SP1/SP2/SP3 must never collapse to one orbit) was correct throughout and untouched.
- **Unspecified stereo is never auto-assigned**: `untagged_tetrahedral_center_is_never_auto_promoted` (new, mirrors the pre-existing `untagged_four_coordinate_pt_is_never_auto_promoted` for square-planar) plus the structural guarantee — `corrected_chirality`'s first line (`if stored == Chirality::None { return Chirality::None; }`, unchanged) means the new module is never even consulted for an atom with no declared chirality.
- Atom-renumbering invariance: `renumber_preserves_rotation_equivalence` — for every rotation in both groups, applying the same (deliberately non-monotonic) bijective id map to two rotation-equivalent configurations keeps them rotation-equivalent.

## Bug fix: square-planar routing into tetrahedral checks (`chematic-3d`)

`tetrahedral_constraint_for`'s only guard was `debug_assert!(atom.chirality != Chirality::None)` — a release-mode no-op that's also wrong even in debug, since `Chirality::SquarePlanar` is also `!= None`. `requires_negative_volume: atom.chirality == Chirality::CounterClockwise` was then always `false` for a `SquarePlanar` atom, silently treating it as a declared `@@` tetrahedral center and deciding Satisfied/Violated by `VOLUME_EPS = 1e-6` floating-point noise — reachable via ordinary (non-dative) square-planar SMILES like `[Pt@SP1](Cl)(Cl)(N)N`, since plain SMILES bonds parse as `BondOrder::Single` and pass this function's own dative-bond exemption.

Fixed with one guard reusing the exact `UnsupportedCoordination` variant the function already returns a few lines below: `if !atom.chirality.is_tetrahedral() { return Err(UnsupportedCoordination); }`.

**Full independent-assertion-chain regression test**, not just the `Unevaluable` verdict in isolation:
1. `tetrahedral_constraint_for` itself rejects the atom — not even a tetrahedral *candidate*.
2. `verify_stereo`'s public report agrees (never `Satisfied`/`Violated`).
3. An independent crate/code path, `chematic_chem::assign_cip`, independently agrees: no CIP code at all — cross-referencing (by name, in the test's own doc comment) the two pre-existing PR #313 regression tests that already prove this for the CIP layer alone: `chematic-chem/src/cip.rs`'s `square_planar_center_never_gets_a_bogus_tetrahedral_cip_code` and `chematic-cip/src/tests.rs`'s `square_planar_center_is_skipped_not_assigned_a_bogus_cip_code` (both re-run and confirmed passing unchanged in this PR).
4. The atom's stored `Chirality::SquarePlanar` value is asserted byte-for-byte unchanged after all of the above — true by construction (every function called takes `&Molecule`, an immutable reference, so mutation is ruled out by the borrow checker, not just by current implementation), asserted so the guarantee is visible and regression-checked.

Coordinates are deliberately puckered (asymmetric, non-mirror-symmetric z-offsets) with the resulting signed-volume magnitude asserted `> VOLUME_EPS` directly in the test, so it's proven the pre-fix code would have reached a confident `Satisfied`/`Violated` verdict here — not degraded to `DegenerateGeometry` for an unrelated reason. Verified empirically: temporarily disabling the guard makes this test fail with `Violated`, confirming it's load-bearing.

## Review round 2: public-API hardening (added after independent review)

An independent review of the first implementation approved the group theory, the SMILES integration, and both bug fixes unchanged, but found two real design gaps in the public API surface that would have been costly to fix after release. Both are now fixed:

**Blocker 1 — arity lock-in.** `StereoGeometry` was `#[non_exhaustive]`, but every other public type/function (`StereoConfiguration`, `CanonicalStereoConfiguration`, `canonicalize_configuration`, `representative()`) was hardcoded to `[u32; 4]`. Adding 5-slot TBP or 6-slot Octahedral later would have forced a breaking signature change to all of them despite the enum itself claiming forward-compatibility.
**Fix (Option A, the more conservative of two offered)**: `StereoConfiguration`, `CanonicalStereoConfiguration`, `canonicalize_configuration`, and `equivalent_under_rotation` are now `pub(crate)` — not part of this crate's public API at all. Only `remap_tetrahedral_parity`, `remap_square_planar_tag`, `StereoGeometry`, and `StereoGeometryError` are public, matching what existing callers (`chematic-smiles`) actually need. This defers the real arity-generalization API design question until a second geometry family genuinely needs it, rather than committing to "every geometry has 4 slots" in public API today.

**Blocker 2 — invalid states were constructible and one function returned a confidently-wrong answer for malformed input.** Three specific gaps, all fixed with new tests:
1. `StereoConfiguration`'s fields were both `pub`, so `StereoConfiguration { geometry: ..., slots: [1, 1, 2, 3] }` (a duplicate slot id) could be constructed directly, bypassing validation. Fixed: fields are now private; the only constructor is `StereoConfiguration::new(geometry, slots) -> Result<Self, StereoGeometryError>`, which runs the same duplicate check `canonicalize_configuration` already ran.
2. `renumber()` didn't reject a non-injective `id_map` — e.g. mapping two distinct input ids to the same output id silently produced a duplicate that could never have been constructed directly. Fixed: `renumber()` now routes its output through `StereoConfiguration::new`, so a renumbering-induced duplicate is caught by the same check, reusing `DuplicateSlotId`.
3. **The real bug**: `remap_tetrahedral_parity(original, canonical)` never checked that `original` and `canonical` named the same 4 ligand ids. Concretely, `remap_tetrahedral_parity([1,2,3,4], [1,2,3,5])` — two arrays with *different* id sets — returned `Ok(true)`, because their (necessarily different) canonical representatives compared unequal, which the function misread as "parity flip needed." A confident-looking wrong answer for a meaningless comparison. Fixed: both arrays are now sorted and compared for exact set equality before any canonicalization runs; a mismatch returns the new `StereoGeometryError::MismatchedLigandSet` variant instead. This case is unreachable from any real parsed molecule today (`original`/`canonical` always share the same underlying atom ids in production use, since both are neighbor orderings of the *same* stereocenter), so the fix changes no existing behavior — it closes a latent public-API correctness gap.

**Downstream call-site fix**: `canonical.rs`'s tetrahedral bridge call previously did `remap_tetrahedral_parity(o, c).unwrap_or(false)`, silently swallowing *any* error (including the newly-possible `MismatchedLigandSet`) into "no parity flip." This is now handled explicitly rather than blanket-swallowed, consistent with the module's own "no silent fallback on invalid input" design principle stated throughout the RFC.

## Review round 2: base-sync fix (found independently, not part of the requested-changes list)

While preparing this PR for merge, I found the branch was still based on `main` at `494d634` — before PR #325 (the `pipeline_v2` vs RDKit re-measurement) merged. GitHub's own `mergeable`/`mergeable_state` fields confirmed this concretely: `false`/`dirty`. A same-file `CHANGELOG.md` insertion-point conflict (both PRs add a section immediately after `## [Unreleased]`) meant a naive merge risked silently dropping PR #325's already-merged measurement entry. Resolved by merging `origin/main` into this branch and combining both `CHANGELOG.md` sections (this PR's 3 entries, followed by PR #325's `### Measured` entry — both fully present in the current file, verified). Re-ran the affected crates' build + tests after the merge before pushing.

**Local test note**: 3 local test failures surfaced immediately after this merge (`ez_bounds_29_corpus_regression_but2ene_z_fixed_nothing_else_broken`, `timeout_failure_still_carries_evidence_computed_before_it_tripped`, `timeout_zero_fails_closed_with_typed_timeout`) — none in files this PR touches. Cross-checked by running the identical 3 tests against a clean `origin/main` checkout (commit `19f7191`, unrelated to this PR) on the same machine: **identical 3 failures**, while an unrelated CPU-heavy process (`renkin`, a different repo's benchmark, ~330% CPU) was concurrently running locally (`uptime` load average 9.90 against 10 physical cores). Confirmed as pre-existing local-machine environmental flakiness, not a regression from this PR or from the base-sync merge — no fixture was excluded, no timeout was loosened, and this PR's own CI (a clean, dedicated runner) passed these same tests cleanly both before and after the merge.

## RDKit oracle/fixture provenance

- Local RDKit checked fresh for this PR: `2026.03.4` (`.venv/bin/python -c "import rdkit; print(rdkit.__version__)"`) — a newer patch than the `2026.03.3` the original square-planar RFC pinned when `SquarePlanarPermutation::trans_pairs()`/`remap_square_planar` were first oracle-derived.
- **No new RDKit oracle calls were made by this PR.** The 144-case oracle table (`scripts/square_planar_permutation_oracle.py`) is reused strictly as a **regression** fixture proving the new rotation-group-based code reproduces the previously-validated `trans_pairs()` semantics — not a fresh oracle validation.

## Performance — stated honestly as inconclusive, no percentage claimed

**No clear performance degradation was observed; the measured difference is within noise and cannot be statistically confirmed as either an improvement or a regression on this shared, noisy machine.**

Benchmark: `chematic-smiles/benches/parse_bench.rs`'s pre-existing `canonical_smiles_10mol` (includes glucose, 5 tetrahedral stereocenters). 6 runs taken across this session (criterion 3-point estimates): before-side (pristine `main`, taken twice) medians 134.53 µs and 124.60 µs; after-side (this PR's code, taken 4 times, one excluded as contaminated by a concurrently-running unrelated `cargo test`) medians 108.40, 104.03, 138.85 µs. `uptime` during these measurements reported **load averages of 14–17** on a shared machine with multiple logged-in users — the same class of contention independently reconfirmed during the base-sync merge above. The before/after absolute-time envelopes overlap substantially (roughly 98–152 µs on both sides), so no percentage change is reported as fact. Architecturally (not empirically): the new path removes a `HashMap<T, usize>` allocation from the common 4-element write path in favor of a fixed-size array scan over a `const` table — a reasonable basis to *expect* a small win, but not one this measurement can confirm. Full 6-run table and reasoning in RFC §13.

## Known, unfixed gaps (out of scope for this PR, documented in RFC §11)

- `chematic-mol/src/mol2000.rs`'s `wedge_vs_3d_conflicts` gates on `Chirality::None` equality rather than `!is_tetrahedral()` — the same gap `docs/rfcs/square_planar_stereo_rfc.md` already fixed at 3 other call sites. Confirmed **unreachable today**: chematic-mol's MOL/SDF readers never produce `Chirality::SquarePlanar`.
- `chematic-core/src/molecule.rs`'s `Molecule::fragments()` silently drops `stereo_neighbor_order` when splitting into connected components — real, pre-existing, unrelated data-loss bug.

## Verification gate — all items re-run at the final head (`cd24b27`) after the base-sync merge

- `cargo build --release` for the 3 affected crates (`chematic-core`, `chematic-smiles`, `chematic-3d`): **PASS**.
- `cargo test --release -p chematic-core -p chematic-smiles -p chematic-3d --all-features`: **534 passed, 3 failed** — the 3 failures are the pre-existing local-environment flakiness documented above (identically reproduced on a clean, unrelated `origin/main` checkout under the same concurrent-load conditions), not caused by this PR.
- `cargo fmt --all -- --check`: **PASS**.
- `cargo clippy` (pre-commit hook, run on every commit including the merge): **PASS**, 0 warnings.
- **CI on the final head (`cd24b27`), a clean dedicated runner**: all 18 checks green — `Test`, `Rust Criterion regression gate`, `Clippy`/`Clippy Lints`, `Format Check`, `Test (WASM)`, `Test (Python bindings)`, `Test (native-inchi)`, `Build Check`, `Cargo Audit`, `Cargo Deny`, `CodeQL`, `Python bench regression gate`, `Validate CITATION.cff`, and 4 `Analyze (*)` jobs. Confirms the 3 local failures above are environmental, not real — a clean runner has no `renkin`-style contention and passed cleanly.
- `grep -rn "vita"` across every file this PR touches or adds: **0 real hits**, re-confirmed at the final head.
- 88-fixture byte-identical canonical-SMILES diff: **PASS**, re-confirmed at the final head.
- Two pre-existing CIP regression tests: **PASS**, unchanged.
- Flagship orbit-count test: **exactly 2** orbits for Tetrahedral, **exactly 3** for Square-planar, over all 24 enumerated (not sampled) permutations of 4 distinct ligands.
- `mergeable: true`, `mergeable_state: clean` against current `main` (confirmed after the base-sync merge, previously `false`/`dirty`).

## Commits (7, in review order)

1. `feat(core): add generalized stereo-configuration geometry foundation` — new module + wiring + RFC, self-contained, zero existing-behavior change.
2. `fix(3d,smiles): stop routing square-planar centers through tetrahedral checks; fix allene-end parity fallback` — the two independent routing-correctness bugs, each with its own regression test.
3. `feat(smiles): wire square-planar canonicalization through the generalized stereo-geometry module` — pure adapter wiring, no bug fix bundled.
4. `docs: avoid self-referential vita-grep false positive in RFC wording`
5. `style: cargo fmt stereo_constraints.rs`
6. Review-round-2 commit — Blocker 1 (`pub(crate)` restriction, Option A), Blocker 2 (validated constructor, `renumber()` duplicate rejection, `MismatchedLigandSet`, call-site fix), corrected cis/trans fixture comment, CHANGELOG entry.
7. `Merge remote-tracking branch 'origin/main'` — base-sync fix, `CHANGELOG.md` conflict resolved preserving both this PR's entries and PR #325's `### Measured` entry.

**Status**: not marked ready for review, not merged. No version bump, no tag, no release, no publish. CI green at `cd24b27`; awaiting explicit merge instruction.


kent-tokyo:
## Review response (commit 89d65ed)

Addressed independent review's 2 blockers + 2 minor issues. Group theory, SMILES integration, and the two prior bug fixes are unchanged.

**Blocker 1 (API arity commitment) — Option A chosen**: `StereoConfiguration`, `CanonicalStereoConfiguration`, `canonicalize_configuration`, `equivalent_under_rotation` are now `pub(crate)`, not `pub`. Only `remap_tetrahedral_parity`/`remap_square_planar_tag` (what `canonical.rs` actually calls) stay public. Chosen over the enum-of-per-geometry-arrays alternative (Option B) as the smaller, more conservative change — defers the real arity-generalization decision entirely rather than guessing at a shape now. RFC §9.0 documents the reasoning.

**Blocker 2 (3 validation gaps), all fixed with dedicated tests**:
1. `StereoConfiguration::new(geometry, slots) -> Result<Self, StereoGeometryError>` is now the only constructor (private fields) — `stereo_configuration_new_rejects_duplicate_slot_id`.
2. `renumber()` now rejects a non-injective `id_map` that manufactures a duplicate — `renumber_rejects_id_map_that_creates_a_duplicate`.
3. **Real bug caught by review**: `remap_tetrahedral_parity([1,2,3,4],[1,2,3,5])` returned `Ok(true)` — a confident wrong answer, since two arrays naming different id sets necessarily canonicalize to unequal representatives, misread by the bare `!=` check as "needs a flip." Fixed with new `StereoGeometryError::MismatchedLigandSet` — `remap_tetrahedral_parity_fails_closed_on_mismatched_ligand_set`. Confirmed `remap_square_planar_tag` does NOT have the analogous gap (it compares via full equality, not a boolean test — already covered by `remap_square_planar_tag_none_on_mismatched_id_set`). The `canonical.rs` call site's `.unwrap_or(false)` is now documented in place as the deliberate 2-state-tag fallback (matching this function's existing philosophy a few lines above), not a silently-swallowed new error class.

**Minor (cis/trans comment)**: the `duplicate_chemistry_distinct_ids_...` test's comment had the chemistry backwards for its chosen slot order (claimed "SP1 = cis" when SP1 there is actually trans-shaped). Fixed by dropping the specific cis/trans claim (renamed to generic "SP1-shaped"/"SP2-shaped" language) rather than risk a second wrong derivation — audited the RFC and `square_planar_stereo.rs`'s own cisplatin/transplatin claims too; those cite a different, unchanged, pre-existing oracle-verified fixture where the label is correct (verified by hand).

**Minor (CHANGELOG)**: added under `[Unreleased]`.

**Re-verified after all fixes**: 88-fixture byte-identical canonical-SMILES diff still holds (exit 0), `cargo test --workspace --all-features` (0 failures, 77 blocks), `--no-default-features` (0 failures, 76 blocks), `cargo clippy --all-targets --all-features -D warnings` clean, `cargo fmt --all -- --check` clean.

Still draft.

PR #353, opened 2026-08-20
docs: refresh stale WASM bundle size (719 KB gzip -> ~1.1 MB gzip)

kent-tokyo:
## Summary

Follow-up to PR #352 (playground's own stale numbers). While fixing the demo, discovered the docs' "719 KB gzip" WASM bundle-size claim — cited 15 times across `README.md` and 5 `docs/` pages — is also stale.

Directly re-measured with the exact CI recipe (`wasm-pack build --release` + `wasm-opt -O3`, matching `.github/workflows/pages.yml`) against current `main`:
- Raw binary: ~2.9 MB (2,942,069–2,943,754 bytes across two independent local builds, cross-checked against the **live deployed site's** ETag byte count, not just a local build)
- Gzip -9: ~1.1 MB (1,092,248–1,092,311 bytes)

Both figures reproduced consistently within build-nondeterminism noise (<0.1%).

**Root cause** (from `Cargo.toml`'s own comment): `[profile.release]` was switched from `opt-level = "z"` (size) to `opt-level = 3` (speed) at some point — very likely most of the ~1.5x growth; the rest is feature growth (LAMMPS, Cube, OpenDX, mmCIF, QCSchema, ORCA, etc. all shipped since 719 KB was last measured).

Also updated the derived "~42x smaller than RDKit.js" ratio to "~27x" everywhere it appears (30 MB / 1.1 MB), so the comparison stays internally consistent rather than only fixing the bare size figure. Still a real, substantial win — 27x smaller is not a regression story.

## Files changed

- `README.md` (8 occurrences)
- `docs/benchmark.md` (3)
- `docs/rdkit-comparison.md` (3)
- `docs/rdkit-migration.md` (1)
- `docs/rdkit_cheatsheet.md` (1)
- `docs/index.md` (1)

Docs-only change, no code touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,13 +19,13 @@ Pure Rust · Zero C/C++ · Python · WebAssembly · [Live Demo](https://kent-tok
 | | chematic | RDKit (Python) | RDKit.js (WASM) |
 |---|---|---|---|
 | **Get started** | `pip install chematic` | conda / cmake required | no Python bindings |
-| **Browser bundle** | **719 KB** | not available | ~30 MB (~42× larger) |
+| **Browser bundle** | **~1.1 MB gzip** | not available | ~30 MB (~27× larger) |
 | **Batch fingerprints** | **~78 µs/mol** (2–3× faster) | ~160–235 µs/mol | — |
 | **Memory safety** | compiler-enforced (Rust) | C++ | C++ |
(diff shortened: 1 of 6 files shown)


kent-tokyo:
Superseded by #357, which updates the same claims using newer like-for-like measurements and corrects the RDKit.js reference size.

PR #362, opened 2026-08-22
docs(rfc): Tautomer & Parent Identity Phase 2 (RFC + acceptance fixtures)

kent-tokyo:
## Summary

RFC-stage deliverable for ROADMAP.md Phase 2 (Tautomer & Parent Identity, v0.20.0), following the same discipline as Phase 1 (#360): audit the existing `tautomer.rs`/`standardize.rs` code before designing anything new, ship RFC + acceptance fixtures only, stop for explicit authorization before implementing.

**Confirmed defect:** `canonical_tautomer` is not invariant across aromatic lactam/lactim tautomer pairs. Tested against `main` (99401a6): 2-pyridone, 4-pyridone, cytosine, uracil, a guanine-class purine, and hypoxanthine (6 real molecules, RDKit formula-verified) all produce two different canonical outputs depending on which tautomer was fed in. Root cause: `BondOrderMatch::Double` never matches `BondOrder::Aromatic`, so no 1,3-/1,5-shift rule fires across an aromatic ring bond, and the existing aromatic-shift mechanism only moves H between ring atoms, never to/from an exocyclic substituent. Fix design is in RFC section 4.4 — **not implemented this round.**

**Confirmed non-bug:** no repeat of Phase 1's stereocenter-corruption class. `tautomer.rs`'s `MoleculeBuilder`-rebuilding transforms already correctly call `copy_stereo_groups_from`/`copy_stereo_from`/`copy_bond_directions_from`, pinned by existing regression tests.

**Confirmed gaps:** budget exhaustion (`max_iter`/`max_tautomers`) is silent, already self-documented via an `#[ignore]`d regression test proving real order-dependence on a 25-independent-site molecule; the 5 unwired `StandardizationStep` variants (`NormalizeGroups`/`FragmentParent`/`ChargeParent`/`IsotopeParent`/`StereoParent`) mostly already have correct underlying functions (`normalize_groups`/`neutralize_charges`/`remove_isotopes`/`remove_stereo`/`select_fragment`) — the gap is an explicit "Parent identity" concept and audit trail, not missing transforms; no customization hook exists for tautomer rules/scoring (both private).

## What's in this PR

- `docs/rfcs/tautomer_parent_identity_phase2_rfc.md` — audit findings, `StandardizationLimits`/result-state design (with the `timeout_ms`-vs-determinism tension resolved explicitly), Parent-function design (`fragment_parent`/`charge_parent`/`isotope_parent`/`stereo_parent`/`tautomer_parent`/`super_parent`), the aromatic lactam/lactim fix design, a minimal custom-rule/scoring hook design, open questions.
- `validation/tautomer_parent_identity_phase2_fixtures.jsonl` (24 rows) + `validation/tautomer_parent_identity_phase2_holdout.jsonl` (5 rows, never used to shape the design above).
- `validation/README.md` — new section summarizing the above.

**No changes under `crates/*/src/**` this round** (verified via `git status --short crates/`).

## Test plan

- [x] All fixture chemistry independently verified (formula match via RDKit 2026.03.4 for tautomer pairs; direct execution against `main` via throwaway examples, deleted before commit, for both the confirmed-failing and confirmed-passing cases)
- [x] `git diff --stat crates/*/src` empty
- Implementation (StandardizationLimits, result-state enum, `*_parent` functions, the §4.4 aromatic-shift fix) is deferred to a follow-up round pending explicit go-ahead, matching Phase 1's (#360) rhythm.


kent-tokyo:
Revision round 2A pushed (36fdd96) addressing design review feedback:

- **charge_parent redefined**: was a bare `neutralize_charges` wrapper (multi-fragment output, contradicting the RFC's own Parent definition). Now `fragment_parent` → neutralize that one fragment. `tp2-17` corrected to a single-fragment expected output.
- **super_parent order fixed**: was self-contradictory (stated order vs. justification disagreed). Now fixed and every intermediate result pinned in `tp2-23`, not just the final one.
- **Limits redefined**: `StandardizationLimits` → `TautomerLimits`, scoped to tautomer computation only. `max_restarts` removed (undefined at this level, belongs to a future Normalization-phase type). `max_transforms`/`max_tautomers` counting semantics precisely defined by re-reading the existing loop code. `Canceled` removed from the result-state enum (no cancellation mechanism exists).
- **Typed API**: tuples/`String` reasons → `ParentResult`/`ParentComputationStatus`/`TautomerAuditRecord` structs, atom/bond-level detail instead of booleans.
- **tp2-04 fixed**, and fixing it surfaced a second, distinct confirmed defect: nitroso/oxime (`CCN=O`/`CC=NO`) also fails to converge, for an unrelated reason (RFC §1.6) — the matching rule is `prefer_forward: false`, likely to avoid enolizing plain ketones via the same any-bridge pattern.
- Renamed `metal_adjacent` → `disconnected_metal_ion_interaction` (no fixture has an actual coordinate bond).
- Added 6 negative/metamorphic fixtures (phenol/anisole/aniline/pyridine N-oxide non-regression, idempotence, atom-order reorder) to guard round 2C's aromatic-shift fix against over-firing.

Implementation is now split into rounds 2B (Parent API + budget visibility) → 2C (aromatic lactam/lactim fix) → 2D (scoring customization), each pending separate go-ahead. Still zero `crates/*/src` changes; staying in draft.

PR #371, opened 2026-08-23
A2: Conformer Ensemble Core (embed_pipeline_v2-backed)

kent-tokyo:
## Summary

Track A, item A2 of the OpenEye/materials advantage RFC
(`docs/rfcs/openeye_materials_advantage_rfc.md`). A1 (PR #370)
reconfirmed the RFC's finding: `embed_pipeline_v2` is the modern,
rigorous, single-conformer embedder (typed 12-stage failure enum,
explicit deterministic seed, real MMFF94/UFF policy gating), but the
public multi-conformer API (`conformer_ensemble()`) routes through the
legacy `etkdg` path instead (no seed, no energy ranking, a live MMFF94
soundness defect). A2 was flagged NEEDS-RESEARCH pending confirmation of
the exact wiring shape.

**Answer: a new outer loop, not an in-place extension.**
`embed_pipeline_v2`'s own `max_attempts`/`timeout_ms` already retry the
*same* target conformer on a bad stochastic draw internally — they don't
generate different conformers. This PR adds `embed_ensemble_v2`, which
calls `embed_pipeline_v2` `count` times with deterministically derived
seeds and accumulates into the *existing* `ConformerEnsemble` storage
type — no new ensemble/storage type.

**Rust core only** (`chematic-3d`). No Python/WASM bindings this round —
separate surface, separate review. The legacy `conformer_ensemble()`/
`generate_conformer_ensemble_with_config` API is completely untouched.

## Design decisions

1. **Energy comparability under `Mmff94WithUffFallback`**: individual
   attempts can resolve to different `actual_force_field_used` (MMFF94
   vs UFF — non-comparable energy scales). Kept conformers are grouped
   by that field (first-seen order) and ranked by ascending energy only
   *within* a group; `mixed_force_field` discloses when more than one
   group is present.
2. **No "rank by energy" flag to misconfigure**: energy is `Some(..)`
   whenever the underlying `EnergyReport` isn't `EnergyReport::None`,
   checked off the actual result data, never inferred from the
   requested policy — removes an invalid-state combination entirely
   rather than validating it.
3. **Seed derivation** reuses the crate's own existing
   `derive_attempt_seed` (widened to `pub(crate)`, no public API
   change) rather than inventing a second scheme.
4. **Symmetric RMSD pruning by default** (new
   `ConformerEnsemble::is_duplicate_symmetric`, mirroring the existing
   `is_duplicate` but via this crate's own automorphism-aware
   `rmsd_symmetric`) — plain Kabsch pruning can treat truly-identical
   conformers on molecules with interchangeable substituents (e.g.
   `-CF3`) as distinct. Exposed as a config bool for the cheaper
   fallback.

## Verification

- 8 new unit tests in `ensemble_v2.rs`: determinism (same base seed →
  identical output), non-aliasing (different seeds → different output),
  `count: 0` early return, `ForceFieldPolicy::None` never reports
  energy, every attempt recorded (success or typed failure), zero
  ensemble-timeout fails closed before any attempt, symmetric pruning
  is at least as aggressive as plain Kabsch, kept conformers are
  energy-ranked ascending (verified against an independently
  re-computed shadow set, not read from internal state).
- 2 new unit tests in `conformer.rs` for `is_duplicate_symmetric`.
- Real spot check (not committed) across 4 diverse molecules (aspirin,
  caffeine, cyclododecane macrocycle, ibuprofen) under
  `Mmff94WithUffFallback`: energies vary meaningfully across attempts,
  near-duplicates prune correctly, all succeed including the macrocycle.
- `bash scripts/check.sh` green: fmt, clippy, full workspace test suite
  (~550 `chematic-3d` tests), cargo-deny, publish graph, version
  consistency.
- No heavy full-corpus remeasurement this round (per this project's
  standing convention); a full benchmark comparison against the legacy
  path is future work, not a gate for landing this.

## Not in scope

Python/WASM bindings, rewiring/deprecating the legacy
`conformer_ensemble()` API, A3 (Shape Runtime) and beyond — all await
separate, explicit authorization.

### DIFF
--- a/crates/chematic-3d/src/conformer.rs
+++ b/crates/chematic-3d/src/conformer.rs
@@ -232,6 +232,24 @@ impl ConformerEnsemble {
             .any(|existing| kabsch_rmsd(coords, existing, n) < rmsd_threshold)
     }
 
+    /// Symmetry-aware counterpart to [`is_duplicate`](Self::is_duplicate): `true`
+    /// if `coords` is within `rmsd_threshold` Å of any existing conformer under
+    /// [`rmsd_symmetric`] (automorphism-aware) rather than plain fixed-index
+    /// Kabsch RMSD. Correct on molecules with interchangeable substituents
+    /// (e.g. a terminal `-CF3`, see `symmetric_rmsd_recovers_zero_when_fixed_index_rmsd_is_wrong`
+    /// in this module's tests) where `is_duplicate` can treat truly-identical
+    /// conformers as distinct. O(k) automorphism enumerations where k is the
+    /// current ensemble size — more expensive than `is_duplicate`, used where
+    /// correctness on symmetric molecules matters more than raw speed.
+    pub fn is_duplicate_symmetric(&self, coords: &Coords3D, rmsd_threshold: f64) -> bool {
+        if rmsd_threshold <= 0.0 {
+            return false;
+        }
+        self.conformers
+            .iter()
+            .any(|existing| rmsd_symmetric(&self.mol, coords, existing) < rmsd_threshold)
+    }
+
     /// Mean pairwise USR dissimilarity across all conformers.
     ///
     /// Returns a value in `[0.0, 1.0]`: 0.0 means all conformers are identical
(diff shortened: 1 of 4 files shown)


kent-tokyo:
Addressed the review's hardening feedback in commit \`669342a\`:

1. **Select-before-prune fix (the critical one)**: restructured into two strict phases — generate all attempts with zero pruning, then group by \`actual_force_field_used\`, sort each group by ascending energy, and greedily dedup within that energy order. This now always keeps the lowest-energy representative of each RMSD cluster, independent of generation order. Verified against real chemistry: aspirin's attempt 0 (53.766 kcal/mol) is now correctly pruned in favor of attempt 4's lower-energy (53.765) near-duplicate — previously it would have kept whichever was generated first.
2. **No cross-group pruning**: adopted the recommended "prune only within a group" policy — a UFF-fallback candidate and an MMFF94 candidate that are geometrically near-duplicate are both kept now, never resolved by generation order. Pinned with a dedicated synthetic-data test.
3. **`EnsembleTermination`**: added `Completed`/`TimedOut` plus `requested_count` on the result — no more inferring timeout from `attempts.len()`.
4. **Typed `ConformerDisposition`**: replaced `kept: bool` with `Kept{conformer_index}` / `PrunedAsDuplicate{representative_attempt_index, rmsd, symmetric}`, plus `fallback_reason: Option<ForceFieldBridgeError>` on `ConformerSuccess`. The "nothing is dropped" doc claim is now actually true.
5. **Config validation**: `embed_ensemble_v2` now returns `Result<EnsembleV2Result, EnsembleV2ConfigError>`, rejecting NaN/infinite/negative `rmsd_threshold`; non-finite energy is treated as no-usable-energy rather than sorted.

7 new regression tests (synthetic-data unit tests for the selection logic per your suggested test architecture, since that's more precise and faster than engineering real pipeline scenarios) plus 2 new `conformer.rs` tests for the new `find_duplicate`/`find_duplicate_symmetric` methods (which replaced the bool-only `is_duplicate`/`is_duplicate_symmetric`, now thin wrappers with no behavior change). Full detail in the commit message.

Still draft, still the same 3 `chematic-3d` files, still zero Python/WASM/legacy-API changes, `bash scripts/check.sh` green. Not merging — awaiting your review of the hardened version.

Issue #256, opened 2026-08-07
chematic-3d: generate_coords places every ring before any chain atom, breaking chain-bridged ring islands (bibenzyl)


kent-tokyo:
## Summary

`dg::generate_coords`'s `place_component` (`crates/chematic-3d/src/dg.rs`,
line 195) places every ring in a component via `place_rings` (line 331)
**before** walking any non-ring chain atom via `dfs_place` (line 584).
When two ring "islands" (sharing no atoms, and having no direct bond to
each other) are connected only through one or more not-yet-placed chain
atoms, `place_rings` has no already-placed anchor to detect for the
second ring at all -- it falls through to a fixed offset that guarantees
no atom collision but not a correct bond length once the bridging chain
atom(s) are placed afterward and reach that ring's already-fixed position.

## Measured (not assumed)

Bibenzyl (`c1ccccc1CCc1ccccc1`, PhCH2CH2Ph, 14 heavy atoms):
worst bonded-pair distance **~8.7 Å** (should be a normal aromatic C-C
bond, ~1.4-1.5 Å, at the chain-to-ring junction).

Pinned as an intentionally-broken regression fixture in
`crates/chematic-3d/src/dg.rs`'s
`generate_coords_bibenzyl_chain_bridged_ring_islands_known_broken` test
(added in PR #253) -- the assertion currently checks `worst > 5.0` and is
meant to start FAILING once this is actually fixed, at which point it
should be replaced with a `assert_bonded_pairs_sane`-style sanity check
instead of deleted.

## Root cause

`dfs_place`'s neighbour walk only visits *unplaced* neighbours
(`crates/chematic-3d/src/dg.rs`, `unplaced_neighbours` filter in
`dfs_place`) -- once it reaches an atom that's already `placed` (a ring
atom `place_rings` positioned earlier), it stops without any correction,
regardless of how far off that position actually is relative to the
chain's own newly-computed trajectory.

## Distinct from PR #253's new-island fix

PR #253 fixed the case where a "new island" ring has a REAL DIRECT BOND to
an already-placed atom (biphenyl, terphenyl) -- detectable at
`place_rings` time since the already-placed atom exists. This case is
different: no already-placed atom is reachable from the not-yet-placed
ring at all, because the connection runs through non-ring atoms that
`place_rings` hasn't gotten to yet. A local fix within `place_rings`
cannot see this; it needs the placement order itself to change.

## Suggested fix direction (not implemented, not proven)

[...]

Issue #255, opened 2026-08-07
chematic-3d: place_rings' fused-ring +y anchor can produce distorted fusion-seam bonds (anthracene)


kent-tokyo:
## Summary

`dg::generate_coords`'s ring-fusion "fuse to a previously-placed ring atom"
branch (`place_rings`, `crates/chematic-3d/src/dg.rs`, the
`shared_atoms.len() >= 2` case around line 481) always extends the new
ring's centre in the `+y` direction from the fusion bond's midpoint:

```rust
ring_cy = (p0.y + p1.y) / 2.0 + r;
```

This is correct only when `+y` happens to be the direction away from the
rest of the already-placed structure. For a linear 3-ring fusion
(anthracene), it is not always correct at the second fusion seam, and
produces geometrically distorted bonds.

## Measured (not assumed)

Anthracene (`c1ccc2cc3ccccc3cc2c1`, 14 heavy atoms), `generate_coords`
output, bonded-pair distances:

| bond | length (Å) | expected (aromatic C-C) |
|---|---|---|
| 3-4 | 2.2644 | ~1.40 |
| 5-6 | 2.0365 | ~1.40 |
| 9-10 | 0.8675 | ~1.40 |
| 11-12 | 1.3163 | ~1.40 |

(Remaining 12 bonds in the molecule are within normal range.)

This is **pre-existing since commit `d45f91b`** (PR #253's first commit,
the ring-fusion-order fix that made anthracene's three rings visit in
fusion-consistent order at all) -- it was never checked at the
bonded-pair level before; only `min_pairwise_distance > 0.3` was asserted
in `generate_coords_anthracene_terminal_rings_not_superimposed`, which
this distortion satisfies (nothing is literally coincident) while still
being geometrically wrong. Confirmed independent of PR #253's later
new-island-anchor/ordering-adjacency changes by temporarily disabling
those and reproducing byte-identical output.

## Distinct from the new-island fix in PR #253

PR #253 fixed an analogous "always extend in one fixed direction"
heuristic for the `shared_atoms.is_empty()` ("new island", e.g. biphenyl)
branch -- but that fix does not apply here. This is the OTHER branch
(`shared_atoms.len() >= 2`, used when a ring genuinely shares atoms with
something already placed, i.e. real ring fusion), a structurally
different code path with the same class of directional bug.

## Suggested fix direction (not implemented, not proven)

Same general approach as PR #253's new-island fix: instead of
[...]

PR #386, opened 2026-08-24
feat(3d): connectivity-ordered ring/chain placement engine (issues #256/#255)

kent-tokyo:
## Summary

Phase 1 of `docs/rfcs/dg_connectivity_ordered_placement_rfc.md` (Phase 0 design doc, merged earlier). Adds a new, additive, crate-private `generate_coords_connectivity_ordered` engine to `crates/chematic-3d/src/dg.rs`, following the RFC's own pre-recorded recommendations exactly.

**Problem** (issues #256/#255): today's `generate_coords` places every ring in a component before walking any chain atom, and its ring-fusion seam math uses a fixed `+y` extension from the shared edge's midpoint. Both are broken, measured directly against the RFC's own fixture table:

| Molecule | Defect | Worst bonded-pair distance (old) | Ideal |
|---|---|---|---|
| naphthalene | #255 fusion seam | 2.2644 Å | ~1.40 Å |
| phenanthrene | #255 fusion seam | 3.3856 Å | ~1.40 Å |
| pyrene | #255 fusion seam | 3.8974 Å | ~1.40 Å |
| bibenzyl (bridge 2) | #256 chain-bridged ring islands | 8.7358 Å | ~1.5 Å |
| diphenylmethane/-propane/-butane (bridge 1/3/4) | #256 | 8.07–8.97 Å | ~1.5 Å |

**Fix, structurally**: ported the traversal technique from `chematic-depict/layout.rs`'s proven `grow_layout`/`dfs_zigzag`/`place_ring_anchored` (verified there against naphthalene/phenanthrene/pyrene/anthracene via `mol.depict_data()`, per the RFC's §3b) — a single worklist discovers and places ring systems and chain atoms in true connectivity order. Ring-fusion center is whichever perpendicular-bisector candidate lies farther from the centroid of everything already placed; winding direction comes from the real, measured signed angle between the two shared anchor atoms — never from the ring template's own listed atom order.

Per the RFC's §4 recommendations: every ring system's plane stays parallel to the global XY plane (option (b), "fixed reference normal") at whatever Z height its entry point naturally lands at; chain placement keeps `dfs_place`'s existing bond-angle/dihedral scheme unchanged (no grandparent-referenced dihedral staggering — scoped strictly to the ring-anchor defects).

**`generate_coords` itself is completely untouched** — this is a parallel implementation (RFC Phase 1), not an in-place rewrite. Not wired up as any live caller's implementation; the routing decision (RFC Phase 3: switch `generate_coords` over, or ship as a new API) is deliberately deferred, same as the RFC's own staged plan.

## Verification

- 17 new tests, called directly against the new engine (not through `generate_coords`, since Phase 1 makes no routing change):
  - Every topology in the RFC's own known-broken table: naphthalene, quinoline, phenanthrene, pyrene, and the bridge-length-1/2/3/4 chain series (diphenylmethane/bibenzyl/diphenylpropane/diphenylbutane) — all now `assert_bonded_pairs_sane`.
  - Extra coverage beyond the RFC's own fixture list: anthracene (linear 3-ring fusion).
  - Positive controls: spiro[5.5]undecane, biphenyl/terphenyl (new-island, bond-anchored, not fused), ibuprofen-shaped ring-with-tail-substituent, a pure chain with no ring at all, determinism across repeated runs, and a broad finite/no-NaN sweep.
- `cargo test -p chematic-3d --lib --release`: 593/593 passing, 0 failures.
- `cargo clippy --workspace --exclude chematic-py --all-targets -- -D warnings` and `cargo fmt --check`: clean.
- `chematic-py` excluded from the local build/clippy pass — a pre-existing local pyo3/Python-framework linker failure, confirmed unrelated by reproducing it identically on `main` before this change (`git stash` + rebuild).

## Test plan

- [x] `cargo test -p chematic-3d --lib --release`
- [x] `cargo clippy --workspace --exclude chematic-py --all-targets -- -D warnings`
- [x] `cargo fmt --check`
- [ ] CI (in progress)

### DIFF
--- a/crates/chematic-3d/src/dg.rs
+++ b/crates/chematic-3d/src/dg.rs
@@ -14,7 +14,7 @@
 //! 4. Each component is offset along the X axis to avoid overlap.
 
 use core::f64::consts::PI;
-use std::collections::VecDeque;
+use std::collections::{BTreeMap, HashMap, HashSet, VecDeque};
 
 use chematic_core::{AtomIdx, BondOrder, Molecule};
 use chematic_perception::find_sssr;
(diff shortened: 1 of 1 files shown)


kent-tokyo:
Follow-up commit strengthens the test suite: the original 17 tests checked bond-length sanity only (`assert_bonded_pairs_sane`), which can't distinguish a correct fix from the exact failure mode the RFC records for the *reverted* #255 fix attempt — "two EXACT atom coincidences plus a 3.7 Å stretch" on phenanthrene, "an exact atom coincidence" on pyrene (RFC §2). Correct bond lengths, unrelated atoms overlapping.

Added `min_pairwise_distance > 0.3 Å` to all fixtures that were missing it, plus a new highest-risk fixture (3-phenylpyridine — the topology where the old engine's fixed `+X` extension pointed a new ring straight back into already-placed structure, 0.14 Å measured before that engine's own centroid-outward fix). All pass: 18/18 in the new suite, including every ring-fusion case with the collision check active.

Also caught and fixed a wrong test fixture along the way: the original "pure chain, no ring" test used octane, which turns out to measure 0.1948 Å min pairwise distance under **both** the old and new engines identically (confirmed by direct comparison) — a pre-existing property of `dfs_place`'s own chain geometry for long unbranched chains, unrelated to this PR. Swapped to pentane.

`cargo test -p chematic-3d --lib --release`: 594/594. Clippy/fmt still clean.

PR #570, opened 2026-09-20
docs: record isotope descriptor coverage

kent-tokyo:
Records the merged full RDKit isotope-mass lookup as non-sealed development evidence in the A0 ledger. It explicitly does not change the rejected-candidate boundary or establish a new adoption result.\n\nValidation: python3 scripts/check_release_docs_consistency.py; git diff --check.


kent-tokyo:
Closing this PR because its reused branch contains previously squash-merged commits and is conflicting by history. The same one-line ledger update is recreated cleanly from current main in PR #571.

